PR #6, opened 2026-05-21
Update repo for Botify usage

ewjoachim:


ewjoachim:
Whoops sorry for the noise.

ewjoachim:
(This project is cool but for various reasons we can't use it directly, so we're forking and I did a PR to change a few details, for us. There is no reason to make the fork private so it's public, but I didn't mean to make the PR against your repo. That said, if you want, you may be interested in cherry-picking some commits, such as the "pytest 9.0.3", the py3.14 or the "Conftest: use tmp_path" ones. Let me know if you want me to make a PR for those)

Issue #28, opened 2026-06-10
Model the no-newline-at-EOF marker as a line property, not a free-floating body line


wkentaro:
## Context

Surfaced during review of the fix for #9 (preserve `\ No newline at end of file` markers). The current code is **correct and tested** — this is a maintainability/structure follow-up, not a bug.

## Problem

The #9 fix keeps git's `\ No newline at end of file` marker as a separate element in the diff body list. Because the marker is a sibling element rather than a property of the line it annotates, "skip the marker" logic is re-implemented independently in several places:

- `git_hunk/_hunk.py` — `_advance_offsets` continue-guard, and the context-count filter in `_build_sub_hunks`
- `git_hunk/_lines.py` — the `prev_kept` stateful flag in `_filter_body_lines`
- `git_hunk/_ui.py` — marker special-case in `_print_hunk_diff`

This also requires exporting `NO_NEWLINE_MARKER` and `is_no_newline_marker` from `_hunk.py` into the other modules.

## Proposal

Parse each body line into a small typed struct (e.g. `BodyLine(text: str, no_newline: bool = False)`) so the marker is structurally attached to the line it annotates. This eliminates the parallel-element protocol, the `prev_kept` flag, and the scattered predicate calls.

## Constraint

`hunk.diff` must still serialize back to raw git patch format (markers emitted as bare `\ No newline at end of file` lines) so `git apply` keeps working.

wkentaro:
> *This was generated by AI during triage.*

## Agent Brief

**Category:** task
**Summary:** Model the `\ No newline at end of file` marker as a per-line property in the structured body, per **ADR 0001 (schema v2)** — `docs/adr/0001-json-v2-hunk-schema.md` (PR #66).

**Current behavior:**
The no-newline marker is a free-floating body line (`\ No newline at end of file`), handled ad hoc in rendering rather than modeled as data.

**Desired behavior (ADR §6):**
In the `show --json` `lines[]` body, a line that lacks a trailing newline carries an optional **`no_newline: true`** property. The marker itself is **not** a separate `lines[]` entry and does **not** consume an `n` index — matching `_print_hunk_diff`, which already excludes the marker from line numbering.

**Acceptance criteria:**
- [ ] A hunk whose last added/removed line has no trailing newline emits that line with `no_newline: true`; no separate marker entry appears in `lines[]`.
- [ ] `n` numbering is unaffected by the marker (verify against `-l`).
[...]

PR #29, opened 2026-06-10
fix: preserve no-newline-at-EOF markers through parse and patch build

### LINKED ISSUE #9: Preserve no-newline-at-EOF markers through parse and patch build
## What to build

Make stage, unstage, and discard work correctly on files whose last line has no
trailing newline. Today every `\ No newline at end of file` marker is dropped
while parsing the diff, so the patch git-hunk rebuilds claims the file ends in a
newline. Two distinct failures result:

1. Editing the last line of a file that has no trailing newline fails outright
   with `patch does not apply`.
2. Changing a file from "ends with newline" to "no trailing newline" reports
   success but silently stages nothing (the staged index is unchanged).

The fix is to preserve the no-newline markers end-to-end (parse → hunk diff →
built patch) and treat them as non-change, non-context lines so they don't
corrupt addition/deletion counts, hunk-header line counts, or split offsets.
Do not append a trailing newline after a hunk that ends in a no-newline marker.

## Acceptance criteria

- [ ] `git-hunk stage <id>` succeeds when the hunk touches the last line of a
      file with no trailing newline, and the staged content matches the working
      tree byte-for-byte (no spurious `\n` added).
- [ ] Staging a newline→no-newline transition actually removes the trailing
      newline from the index (verify with `git show :file | xxd`).
- [ ] unstage and discard round-trip correctly for the same files.
- [ ] Addition/deletion counts in `list` are unchanged by the marker lines.
- [ ] e2e tests cover: edit-last-line-no-newline, newline→no-newline,
      no-newline→newline.

Repro (currently fails):

    printf 'a\nb\nc' > f.txt   # no trailing newline
    git add f.txt && git commit -m init
    printf 'a\nb\ncX' > f.txt
    git-hunk stage <id>        # -> error: patch does not apply

## Blocked by

None - can start immediately.

wkentaro:
Fixes #9.

Every `\ No newline at end of file` marker was dropped while parsing the diff, so the patch git-hunk rebuilt always claimed the file ended in a newline. Staging the last line of a no-newline file failed with `patch does not apply`, and a newline -> no-newline change reported success while silently staging nothing.

This preserves the markers end-to-end (parse -> hunk split -> built patch) and treats them as non-change, non-context lines so they never corrupt addition/deletion counts, hunk-header line counts, or split offsets. `show` line numbers and `-l` line selection skip markers consistently, and a marker is dropped only when its preceding line is.

A representational follow-up (model the marker as a property of its line rather than a free-floating body line) is tracked in #28.

## Test plan
- [x] e2e covers edit-last-line-no-newline, newline -> no-newline, no-newline -> newline, plus unstage/discard round-trips and `-l` selection: `tests/e2e/no_newline_test.py`
- [x] unit covers marker preservation, counts, and split offsets: `tests/unit/_hunk/no_newline_test.py`
- [x] `make lint` and `uv run pytest` (105 passed)

### DIFF
--- a/git_hunk/_hunk.py
+++ b/git_hunk/_hunk.py
@@ -4,7 +4,14 @@ from collections import Counter
 from dataclasses import dataclass
 from dataclasses import replace
 from typing import Any
+from typing import Final
 from typing import TypedDict
+
+NO_NEWLINE_MARKER: Final = "\\ No newline at end of file"
+
+
+def is_no_newline_marker(line: str) -> bool:
+    return line == NO_NEWLINE_MARKER
 
 
 @dataclass(frozen=True)
@@ -101,6 +108,8 @@ def _advance_offsets(
 ) -> tuple[int, int]:
     for j in range(start, end):
         line = body_lines[j]
+        if is_no_newline_marker(line):
+            continue
         if line.startswith("+"):
             new += 1
         elif line.startswith("-"):
@@ -149,7 +158,11 @@ def _build_sub_hunks(
 
         sub_body = body_lines[body_start:body_end]
         additions, deletions = count_changes(sub_body)
-        ctx = sum(1 for line in sub_body if not _is_change(line))
+        ctx = sum(
+            1
+            for line in sub_body
+            if not _is_change(line) and not is_no_newline_marker(line)
+        )
         old_count = ctx + deletions
         new_count = ctx + additions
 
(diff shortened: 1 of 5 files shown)


wkentaro:

The bug was `parse_diff` stripping every `\ No newline at end of file` line, so `build_patch` emitted patches that lied to `git apply` about trailing-newline state. The fix preserves markers in `hunk.diff`, has `_advance_offsets` and the context counter skip them so `@@` line numbers are not inflated, and `_filter_body_lines` ties each marker to its preceding line via `prev_kept` so `-l` selection neither drops a needed marker nor keeps an orphan. All three edge cases (old-only, new-only, both) are covered; the unit test asserting the exact `@@ -8,4 +8,4 @@` header would fail against the pre-fix code. No blocking issues.

Issue #30, opened 2026-06-10
Parse paths git always quotes (tab, newline, backslash, quote in filename)


wkentaro:
## Context

Surfaced during review of #10. That fix uses `core.quotePath=false` to stop git octal-escaping non-ASCII paths. However, git **always** double-quotes and C-escapes filenames containing a tab, newline, backslash, or double-quote — regardless of `core.quotePath` (see `git config` docs for `core.quotePath`).

## Problem

For such a path the diff header is `diff --git "a/od\td" "b/od\td"` (quoted form). `extract_file_path` anchors on the unquoted `a/...` and returns `None`, so the file is **silently dropped** from `list`/`stage`. The quoted form from `git ls-files` (without `-z`) has the same issue, though #10's `-z` switch already protects the untracked path.

## Proposal

Add a quoted-form branch to `extract_file_path`: match `diff --git "a/(.+)" "b/(.+)"$` and C-unescape the captured path (handle `\t \n \\ \"` and octal `\NNN`). Alternatively, enumerate changed paths via `git diff --name-only -z` and fetch each file's diff with `git diff -- <path>`, which sidesteps quoting entirely (the approach GitPython uses).

## Scope notes

- Distinct from #11 (decode raw bytes with surrogateescape) and #10 (non-ASCII octal escaping).
- Filenames with embedded tabs/newlines are rare; low priority, but it is a real silent-drop correctness gap.

wkentaro:
> *This was generated by AI during triage.*

## Agent Brief

**Category:** bug
**Summary:** Files whose paths git quotes (tab, newline, backslash, double-quote in the name) are silently dropped from `list`/`stage`/`discard`. Teach the diff-header path extractor to parse git's quoted, C-escaped header form.

**Current behavior:**
git **always** double-quotes and C-escapes a path containing a tab, newline, backslash, or double-quote — regardless of `core.quotePath` (which the tool already sets to `false`; that only governs octal-escaping of non-ASCII bytes). For such a path the per-file diff header is the quoted form, e.g.:

```
diff --git "a/od\ttab.txt" "b/od\ttab.txt"
diff --git "a/od\\back.txt" "b/od\\back.txt"
```

[...]

Issue #11, opened 2026-06-09
Decode git output with surrogateescape to avoid UnicodeDecodeError crash


wkentaro:
## What to build

Stop git-hunk from crashing with an uncaught `UnicodeDecodeError` traceback when
a diff contains bytes that are not valid UTF-8 (e.g. a Latin-1 source file).
Decode (and encode) git's stdin/stdout/stderr with `errors="surrogateescape"`
so undecodable bytes round-trip losslessly back to git when patches are applied.

## Acceptance criteria

- [ ] `git-hunk list` on a repo containing a tracked file with non-UTF-8 bytes
      exits 0 and lists the hunk (no traceback).
- [ ] Staging such a hunk applies the original bytes unchanged.
- [ ] e2e test covers a repo with non-UTF-8 file content.

## Blocked by

None - can start immediately.

PR #31, opened 2026-06-10
fix: parse git's quoted and escaped file paths

### LINKED ISSUE #10: Parse git's quoted/escaped file paths (diff headers and ls-files)
## What to build

Parse the file paths git emits even when they are quoted and octal-escaped.
With git's default `core.quotePath=true`, a diff header for a non-ASCII path
looks like `diff --git "a/\321\204..." "b/\321\204..."`. git-hunk's path regex
expects a bare `a/...`, fails to match, and silently drops the whole file from
`list`/`stage`. The same quoting affects untracked paths from
`git ls-files --others`, which currently land in output with literal quotes and
escapes.

Make path handling robust across all entry points: the diff-header parse, the
patch file-header lookup, and untracked-file listing. Either force
`core.quotePath=false` (and use `-z` where applicable) for git invocations, or
unquote git's C-style quoting. Also fix the diff-header regex so a filename that
itself contains the substring ` b/` is parsed correctly (the non-greedy match
currently splits at the wrong point).

## Acceptance criteria

- [ ] A modified file with a non-ASCII name appears in `list` and can be
      staged/unstaged/discarded, with default git config (quotePath unset).
- [ ] An untracked file with a non-ASCII name shows its real path (no quotes or
      `\NNN` escapes) in `list` and `list --json`.
- [ ] A filename containing the literal substring ` b/` is parsed to the correct
      path.
- [ ] e2e tests cover non-ASCII modified, non-ASCII untracked, and the ` b/`
      edge case.

## Blocked by

None - can start immediately.

wkentaro:
Fixes #10.

With git's default `core.quotePath=true`, non-ASCII file paths are octal-escaped and double-quoted in diff headers (`diff --git "a/\321\204..." "b/..."`) and in `git ls-files` output. git-hunk's path regex expected a bare `a/...`, so it failed to match and silently dropped the whole file from `list`/`stage`. A filename containing the literal substring ` b/` was also mis-split by the non-greedy regex.

## Summary
- Force `core.quotePath=false` on every git invocation (via `run_git`) so paths arrive literally, and read untracked paths NUL-delimited (`-z`).
- Add a shared `extract_file_path` helper used by both `parse_diff` and the patch builder. It anchors on the fact that the two halves of `diff --git a/<p> b/<p>` are identical for non-renames (backreference `a/(.+) b/\1$`), so a path containing ` b/` splits correctly; a non-greedy fallback preserves prior behavior for the rename case.
- Route `is_git_repo` through `run_git` so all git calls share one invocation path.

## Known limitations / follow-ups
- Non-UTF-8 byte paths still need surrogateescape decode/encode — tracked in #11.
- Filenames containing tab/newline/backslash/quote are always quoted by git regardless of `core.quotePath`, so they remain unparsed — tracked in #30.

## Test plan
- [x] e2e: non-ASCII modified file round-trips (stage/unstage/discard), non-ASCII untracked shows its real path in `list`/`--json`, and a ` b/`-containing path round-trips: `tests/e2e/quoted_paths_test.py`
- [x] unit: `extract_file_path` for plain, non-ASCII, and ` b/` paths: `tests/unit/_hunk/extract_file_path_test.py`

### DIFF
--- a/git_hunk/_git.py
+++ b/git_hunk/_git.py
@@ -2,17 +2,12 @@ import subprocess
 
 
 def is_git_repo() -> bool:
-    result = subprocess.run(
-        ["git", "rev-parse", "--is-inside-work-tree"],
-        capture_output=True,
-        text=True,
-    )
-    return result.stdout.strip() == "true"
+    return run_git("rev-parse", "--is-inside-work-tree", check=False).strip() == "true"
 
 
 def run_git(*args: str, input: str | None = None, check: bool = True) -> str:
     result = subprocess.run(
-        ["git"] + list(args),
+        ["git", "-c", "core.quotePath=false"] + list(args),
         capture_output=True,
         input=input.encode() if input is not None else None,
     )
@@ -34,8 +29,8 @@ def get_diff(staged: bool = False, files: list[str] | None = None) -> str:
 
 
 def get_untracked_files() -> list[str]:
-    output = run_git("ls-files", "--others", "--exclude-standard")
-    return [f for f in output.strip().split("\n") if f]
+    output = run_git("ls-files", "--others", "--exclude-standard", "-z")
+    return [f for f in output.split("\0") if f]
 
 
 def apply_patch(patch: str, *, cached: bool = False, reverse: bool = False) -> None:
--- a/git_hunk/_hunk.py
+++ b/git_hunk/_hunk.py
@@ -182,6 +182,17 @@ def _split_hunk(
     return _build_sub_hunks(header_line, body_lines, regions, split_points)
 
 
+def extract_file_path(file_diff: str) -> str | None:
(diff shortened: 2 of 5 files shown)


wkentaro:

Solves the root cause by forcing `core.quotePath=false` on every git invocation, so non-ASCII paths arrive as raw UTF-8 and no octal/C-escape decoding is needed in the Python layer. `extract_file_path` resolves the ` b/` ambiguity with a backreference regex (non-rename headers have identical `a/`//`b/` halves) and a non-greedy fallback for renames, and untracked listing switches to NUL-delimited `-z` output. e2e + unit tests cover non-ASCII modified, non-ASCII untracked, and a ` b/` filename. The title mentions "escaped paths" but the approach sidesteps escaping entirely; the residual git-C-escaped cases are honestly tracked in #30. No blocking issues.

Issue #23, opened 2026-06-09
Document and version the --json output schema


wkentaro:
## What to build

Give `--json` a documented, versioned contract so downstream agents and tools
can depend on it. Real usage shows agents guessing at the shape (a flat list of
hunk objects vs. a nested by-file/by-status structure) and silently failing when
they guess wrong. Pin the schema down:

- Add a `schema_version` to the JSON envelope.
- Document the exact shape (fields, types, status values) in the README.
- Optionally expose the JSON Schema (e.g. a `--json-schema` flag) so consumers
  can validate.

## Acceptance criteria

- [ ] `list --json` output carries an explicit `schema_version`.
- [ ] The JSON shape is documented in the README with field descriptions.
- [ ] The contract is covered by a test that would fail if the shape changes
      unintentionally.

## Blocked by

None - can start immediately.

PR #32, opened 2026-06-10
fix: handle non-UTF-8 bytes in git output without crashing

### LINKED ISSUE #11: Decode git output with surrogateescape to avoid UnicodeDecodeError crash

wkentaro:
Fixes #11.

A diff containing bytes that are not valid UTF-8 (e.g. a Latin-1 source file) crashed git-hunk with an uncaught `UnicodeDecodeError` traceback at `result.stdout.decode()`.

## Summary
- Decode and encode git's stdin/stdout/stderr with `errors="surrogateescape"` so undecodable bytes round-trip losslessly: the rebuilt patch hands `git apply` exactly the bytes git emitted.
- Hash hunk ids with the same handler so id computation does not crash on the resulting surrogate code points.
- Backslash-escape surrogates for display (`list`, `show`, the applied message, and error output) so they never crash the Rich layer with `UnicodeEncodeError`, while patch building keeps the raw bytes.

## Known limitation
- `list --json` emits surrogates as `\udcXX` escapes (lone surrogates), which Python's `json` round-trips but stricter parsers reject. The right place to settle the JSON byte representation is the schema work in #23, so it's deliberately out of scope here.

## Test plan
- [x] e2e: `list`, `show`, and `stage` on a repo with non-UTF-8 file content exit 0; staging applies the original bytes unchanged (`git show :file` byte-compare): `tests/e2e/non_utf8_test.py`
- [x] unit: `_safe` is identity on valid text and makes surrogate-bearing strings strict-UTF-8 encodable: `tests/unit/_ui/safe_test.py`
- [x] `make lint` and `uv run pytest` (101 passed)

### DIFF
--- a/git_hunk/_git.py
+++ b/git_hunk/_git.py
@@ -6,16 +6,18 @@ def is_git_repo() -> bool:
 
 
 def run_git(*args: str, input: str | None = None, check: bool = True) -> str:
+    # Git output and input may contain bytes that are not valid UTF-8 (e.g. a
+    # Latin-1 source file). surrogateescape round-trips those bytes losslessly
+    # so a rebuilt patch hands git back exactly what it emitted.
     result = subprocess.run(
         ["git", "-c", "core.quotePath=false"] + list(args),
         capture_output=True,
-        input=input.encode() if input is not None else None,
+        input=input.encode(errors="surrogateescape") if input is not None else None,
     )
     if check and result.returncode != 0:
-        raise RuntimeError(
-            f"git {' '.join(args)} failed: {result.stderr.decode().strip()}"
-        )
-    return result.stdout.decode()
+        stderr = result.stderr.decode(errors="surrogateescape").strip()
+        raise RuntimeError(f"git {' '.join(args)} failed: {stderr}")
+    return result.stdout.decode(errors="surrogateescape")
 
 
 def get_diff(staged: bool = False, files: list[str] | None = None) -> str:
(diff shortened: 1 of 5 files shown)


wkentaro:

The decode/encode `surrogateescape` pairing is consistent end-to-end: `run_git` decodes stdout and re-encodes `input` with `errors="surrogateescape"`, and both hash functions in `_hunk.py` match, so bytes round-trip losslessly (verified by the e2e byte-comparison test). The display layer applies `_safe()` (backslashreplace) at every Rich sink where a surrogate-bearing string could appear, preventing `UnicodeEncodeError` on strict stdout. The `--json` lone-surrogate limitation is correctly deferred to #23.

Note: during processing I fixed the failing `lint` job by changing the test literal `b"caf\xe9"` to `b"pass\xe9"` -- the `typos` checker flagged `caf` as a misspelling of `calf`. This keeps the `0xe9`/Latin-1 byte the test exercises and is force-pushed onto the branch; CI is now green. No blocking issues.

wkentaro:
need rebase

PR #33, opened 2026-06-10
fix: apply partial -l selection correctly for unstage and discard

### LINKED ISSUE #12: Fix partial -l selection for unstage/discard on multi-group hunks
## What to build

Make partial line selection (`-l`) correct for unstage and discard when a hunk
contains more than one change group. Today the line filter converts every
unselected `-` line into a context line carrying the OLD content. That works for
a forward apply against the working tree, but for a reverse apply (unstage,
discard) the target already holds the NEW content, so the context line does not
match and git rejects the patch with `patch does not apply`.

Build the filtered patch so that unselected change groups are excluded as
genuinely separate hunks (each with valid surrounding context for the direction
being applied), rather than folded into the same hunk as opposite-side context.
The result must apply cleanly in both directions.

## Acceptance criteria

- [ ] `git-hunk unstage <id> -l <subset>` succeeds and produces the correct
      index content when the hunk has two or more change groups.
- [ ] `git-hunk discard <id> -l <subset>` likewise restores exactly the selected
      lines from HEAD and leaves the rest untouched.
- [ ] A full round-trip (`stage -l A`, then `stage` the remainder, then
      `unstage -l A`, then `discard -l B`) leaves index and working tree in the
      expected state at every step.
- [ ] e2e tests cover multi-group hunks for stage, unstage, and discard with
      both include and exclude (`^`) specs.

Repro (currently fails):

    # file with two changed lines in one hunk
    git-hunk discard <id> -l 2,3   # -> error: patch does not apply

## Blocked by

None - can start immediately.

wkentaro:
Fixes #12.

`git-hunk unstage <id> -l <subset>` and `discard <id> -l <subset>` failed with `patch does not apply` whenever the hunk had more than one change group. The line filter always converted unselected `-` lines into context carrying the OLD content, which only matches a forward (stage) apply; unstage and discard reverse-apply against content that already holds the NEW side, so the context did not match.

## Summary
- Make the line filter direction-aware: a forward apply keeps unselected `-` lines as context and drops `+`, while a reverse apply (unstage, discard) keeps unselected `+` as context and drops `-`.
- Thread a `reverse` flag from `_run_patch_command` through `_apply_line_filter` to `filter_hunk_lines`. Forward (`stage`) behavior is byte-identical to before.

## Test plan
- [x] e2e: multi-group hunk for stage/unstage/discard with both include and exclude (`^`) specs, plus a full `stage -l A` → `stage` remainder → `unstage -l A` → `discard -l A` round-trip: `tests/e2e/line_selection_test.py`
- [x] unit: `filter_hunk_lines(..., reverse=True)` include and exclude keep the NEW side as context and drop the OLD side: `tests/unit/_lines/filter_test.py`

### DIFF
--- a/git_hunk/_cli.py
+++ b/git_hunk/_cli.py
@@ -109,14 +109,16 @@ def _find_hunks_by_ids(hunks: list[Hunk], ids: list[str]) -> list[Hunk]:
     return found
 
 
-def _apply_line_filter(hunks: list[Hunk], line_spec: str | None) -> list[Hunk]:
+def _apply_line_filter(
+    hunks: list[Hunk], line_spec: str | None, *, reverse: bool
+) -> list[Hunk]:
     if line_spec is None:
         return hunks
     if len(hunks) != 1:
         raise CliError("line selection (-l) requires exactly one hunk id")
     try:
         lines, exclude = parse_line_spec(line_spec)
-        return [filter_hunk_lines(hunks[0], lines, exclude=exclude)]
+        return [filter_hunk_lines(hunks[0], lines, exclude=exclude, reverse=reverse)]
     except ValueError as exc:
         raise CliError(str(exc)) from exc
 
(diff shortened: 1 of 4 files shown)


wkentaro:

For a reverse-apply, git matches context lines against the NEW content, so the old code (always keep unselected `-` lines, drop `+` lines) caused context-mismatch failures on unstage/discard. The fix swaps the drop/keep prefixes under `reverse=True` -- keeping unselected `+` lines as context (the side git scans) and dropping `-` lines. I traced all six cases (stage/unstage/discard x include/exclude) and each outcome is correct; the `reverse=False` default preserves `stage` behavior. The two-group fixture is the minimal reproduction, and the unit + e2e tests verify the git index/working-tree state end-to-end. No blocking issues.

wkentaro:
need rebase

PR #35, opened 2026-06-10
fix: harden CLI input validation for empty ids and bad -l specs

### LINKED ISSUE #15: Harden CLI input validation (empty IDs, malformed -l ranges)
## What to build

Harden CLI argument validation against two footguns:

1. An empty-string hunk ID matches every hunk, because prefix matching uses
   `startswith("")` which is always true. `git-hunk discard ""` (e.g. from an
   unset shell variable) then silently operates on a hunk the user never named.
   Reject empty/whitespace-only IDs explicitly.
2. A malformed `-l` range such as `1-2-3` surfaces a raw Python
   `invalid literal for int()` message. Validate the spec and emit a clear,
   actionable error instead.

## Acceptance criteria

- [ ] An empty or whitespace-only hunk ID is rejected with a clear error and
      never matches a hunk.
- [ ] `-l 1-2-3` (and similar malformed specs) produce a readable error naming
      the bad token, not a Python traceback message.
- [ ] Unit/e2e tests cover both cases.

## Blocked by

None - can start immediately.

wkentaro:
Fixes #15.

Two CLI argument footguns:

1. An empty-string hunk id matched **every** hunk, because prefix matching uses `startswith("")` which is always true — so `git-hunk discard ""` (e.g. from an unset shell variable) silently operated on a hunk the user never named.
2. A malformed `-l` range like `1-2-3` surfaced a raw Python `invalid literal for int()` message.

## Summary
- Reject empty/whitespace-only hunk ids in `_find_hunks_by_ids` before the prefix match runs (covers stage/unstage/discard/show).
- Validate `-l` tokens through a `_parse_line_number` helper (strict positive decimal) and require a range to have exactly two non-empty bounds, so malformed specs produce a readable error naming the offending token. Spaces around a hyphen (`2 - 4`) are still accepted.

## Test plan
- [x] unit: malformed `-l` specs (`1-2-3`, `abc`, `1-`, `+5`, bare `^`) error readably; `2 - 4` still parses: `tests/unit/_lines/parse_spec_test.py`
- [x] e2e: `discard ""` and `show ""` are rejected and never match a hunk; `-l 1-2-3` gives a readable error (not a raw `int()` message): `tests/e2e/error_test.py`

### DIFF
--- a/git_hunk/_cli.py
+++ b/git_hunk/_cli.py
@@ -92,6 +92,8 @@ def _get_hunks(staged: bool, files: list[str] | None = None) -> tuple[list[Hunk]
 def _find_hunks_by_ids(hunks: list[Hunk], ids: list[str]) -> list[Hunk]:
     found = []
     for hunk_id in ids:
+        if not hunk_id.strip():
+            raise CliError("hunk id must not be empty or whitespace")
         matches = [h for h in hunks if h.id.startswith(hunk_id)]
         if len(matches) == 0:
             available = [h.id for h in hunks]
--- a/git_hunk/_lines.py
+++ b/git_hunk/_lines.py
@@ -5,6 +5,16 @@ from ._hunk import Hunk
 from ._hunk import count_changes
 
 
+def _parse_line_number(token: str) -> int:
+    token = token.strip()
+    if not re.fullmatch(r"[0-9]+", token):
+        raise ValueError(f"invalid line number: '{token}'")
+    n = int(token)
+    if n < 1:
+        raise ValueError(f"line numbers must be positive: '{token}'")
+    return n
+
+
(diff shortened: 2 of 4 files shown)


wkentaro:

The empty-id guard (`not hunk_id.strip()`) lands before `startswith("")` could ever match-all, closing that silent bug. `_parse_line_number` enforces positive-decimal-only via `re.fullmatch(r"[0-9]+")` plus a `< 1` check, correctly rejecting `abc`, `+5`, `0`, and `00`; range splitting on `"-"` rejects `1-2-3` and `1-` while still allowing the lenient `2 - 4`. Seven unit cases cover all rejection paths plus the allowed-spaces case, and three e2e tests confirm exit code, error text, and -- critically for discard -- that the working tree is left intact. No blocking issues.

Issue #17, opened 2026-06-09
Gate PyPI publish on tests and use full clone for versioning


wkentaro:
## What to build

Make the PyPI publish workflow safe to trigger on a tag:

1. Gate publishing on a passing test run, so a tag with red tests cannot ship a
   broken package.
2. Do a full clone (`fetch-depth: 0`) in the publish job so hatch-vcs derives the
   real version from tags instead of falling back to `0.0.0.dev0`.

## Acceptance criteria

- [ ] The publish job does not run unless tests pass (dependency on the test job
      or an inline test run).
- [ ] The publish checkout uses `fetch-depth: 0`.
- [ ] A dry run / build in CI reports the correct version derived from the tag.

## Blocked by

None - can start immediately.

PR #36, opened 2026-06-10
build: constrain sdist to the package, tests, and metadata

### LINKED ISSUE #16: Stop sdist from shipping unrelated tmp/ files
## What to build

Stop the source distribution from shipping unrelated files. `uv build`
currently produces an sdist containing ~345 files from `tmp/agent-browser/`
(a nested git repo) in addition to the package — that would be published to
PyPI. Constrain the sdist to the package and its support files so the tarball
contains only what belongs in the release.

## Acceptance criteria

- [ ] `tar tzf dist/*.tar.gz` contains no `tmp/` entries and no files outside the
      package, tests, and top-level metadata (README, LICENSE, pyproject).
- [ ] The wheel still contains `git_hunk/skills/core/SKILL.md`.
- [ ] `git-hunk skills get core` works from a clean install built from the sdist.

## Blocked by

None - can start immediately.

wkentaro:
Fixes #16.

`uv build` produced an sdist containing `tmp/` plus every other repo-root file (`docs/`, `assets/`, `Makefile`, `uv.lock`, `.github/`, …) — hatchling includes all non-gitignored files, and `tmp/` is untracked but not ignored, so it (and, historically, a ~345-file nested `tmp/agent-browser/` repo) would be published to PyPI.

## Summary
- Add a `[tool.hatch.build.targets.sdist]` `include` whitelist so the sdist ships only `git_hunk/`, `tests/`, and `README.md`. `pyproject.toml` and `LICENSE` are auto-included by hatchling; `README.md` is **not**, because the readme is dynamic (`hatch-fancy-pypi-readme`) and the sdist→wheel build re-runs that hook — verified empirically.
- The wheel target is untouched, so it still ships `git_hunk/skills/core/SKILL.md`.

## Verification
- [x] `tar tzf dist/*.tar.gz` → zero `tmp/` entries; nothing outside `git_hunk/`, `tests/`, `README.md`, `LICENSE`, `pyproject.toml` (+ hatchling's standard `PKG-INFO`/`.gitignore`)
- [x] wheel still contains `git_hunk/skills/core/SKILL.md`
- [x] clean `pip install` from the built sdist → `git-hunk skills get core` works
- [x] `make lint` and `uv run pytest` (95 passed)

No automated test: the fix is a declarative whitelist (self-guarding — it can only ship what's listed), and a build-shelling test would be slow/flaky; build verification belongs in the publish CI (tracked separately in #17).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -48,6 +48,11 @@ replacement = '![\1](https://raw.githubusercontent.com/wkentaro/git-hunk/main/\2
 [tool.hatch.version]
 source = "vcs"
 
+[tool.hatch.build.targets.sdist]
+# pyproject.toml and LICENSE are auto-included; README.md is not, because the
+# readme is dynamic (hatch-fancy-pypi-readme) and the wheel build needs it.
+include = ["/README.md", "/git_hunk"]
+
 [dependency-groups]
 dev = [
   "mdformat-frontmatter>=2.0.10",


wkentaro:

Adds a `[tool.hatch.build.targets.sdist]` `include` whitelist (`/README.md`, `/git_hunk`, `/tests`). Hatchling auto-includes `pyproject.toml` and `LICENSE` regardless, and `README.md` must be listed explicitly because `hatch-fancy-pypi-readme` renders it at sdist-to-wheel rebuild time; the `git_hunk/skills/.../SKILL.md` data file is retained inside the `/git_hunk` subtree. This is a safe whitelist (can only include what's listed) that resolves the `tmp/` pollution in #16, and the author verified the output with `tar tzf`, a clean `pip install`, and the full test suite. No blocking issues.

wkentaro:
why do we need to include `tests/`? 

Screenshot 2026-06-13 at 17-41-22

wkentaro:

Good question — `tests/` isn't actually required here. Nothing in the build, install, or CI path depends on it: `uv build` produces the sdist but CI runs `uv run pytest tests/` from the repo checkout (not from the sdist), and the wheel never contains `tests/` regardless of this setting.

The reason it's a common convention is downstream re-testing: source-based packagers (Linux distros, conda-forge) build from the sdist and often run the test suite as part of their packaging, so shipping `tests/` lets them verify the build against their own interpreter/dependency versions. It's a low-cost nicety, not a necessity.

Since #16's actual goal was just to stop `tmp/` (and other junk) leaking into the sdist, the minimal whitelist is `/README.md` + `/git_hunk`. If you'd prefer a lean sdist, I can drop `/tests` from the include list; if you want to keep the door open for downstream packagers, leaving it in is fine. Want me to drop it?

wkentaro:
strip it

wkentaro:

Done — stripped `/tests` from the sdist `include`, so the whitelist is now `["/README.md", "/git_hunk"]`. Verified with `uv build --sdist`: the tarball ships `git_hunk/` (including `skills/core/SKILL.md`), `README.md`, `LICENSE`, and `pyproject.toml`, with no `tests/` and no `tmp/`. CI runs tests from the repo checkout, so it's unaffected. Rebased onto current `main` and updated the commit title/body to drop the now-inaccurate "tests" wording.

Issue #37, opened 2026-06-10
Harden the PyPI publish workflow (SHA-pin actions, environment protection, build/publish split)


wkentaro:
## Context

Surfaced during review of #17 (gating publish on tests + full-clone versioning). The gating fix is in, but the publish workflow has further supply-chain hardening opportunities that are out of #17's scope and partly need repo-settings config.

## Proposed hardening

1. **Pin actions to commit SHAs.** `actions/checkout@v4`, `astral-sh/setup-uv@v6`, and especially `pypa/gh-action-pypi-publish@release/v1` (a mutable *branch* pointer) are all mutable refs. Pin each to a full commit SHA with a version comment, and enable Dependabot's `github-actions` ecosystem to keep them current. (Resolve the real current SHAs via the API — do not guess.)
2. **Add environment protection.** Give the `publish` job `environment: name: pypi` and configure the `pypi` environment in repo Settings with required reviewers / tag policy, scoping the PyPI trusted publisher and adding an approval gate. (Requires repo-admin config.)
3. **Separate build from publish.** Per PyPA guidance, split into a `build` job (`uv build` + `upload-artifact`) and a `publish` job (`download-artifact` + publish) so build/test code never holds `id-token: write`.
4. Optional: `persist-credentials: false` on checkout; `SOURCE_DATE_EPOCH` for reproducible builds.

## Scope notes

- Pure hardening; no functional change to what gets published.
- #17 already covers the test gate, `fetch-depth: 0`, and tag-version verification.

wkentaro:
> *This was generated by AI during triage.*

## Agent Brief

**Category:** task
**Summary:** Supply-chain hardening of the PyPI publish workflow — SHA-pin all actions (+ Dependabot), split build from publish so build/test code never holds the OIDC token, and add reproducible-build niceties. No change to what gets published.

**Current behavior:**
The publish workflow (triggered on `v*` tags) runs a single `publish` job that holds `id-token: write`, checks out the repo, runs `uv build`, verifies the built version matches the tag, then publishes. All actions are pinned to **mutable refs** — `actions/checkout@v4`, `astral-sh/setup-uv@v6`, and `pypa/gh-action-pypi-publish@release/v1` (the last is a mutable *branch* pointer). The job that builds and runs project code is the same job that carries the publishing token.

**Desired behavior:**
[...]

PR #38, opened 2026-06-10
ci: gate PyPI publish on tests and use a full clone for versioning

### LINKED ISSUE #17: Gate PyPI publish on tests and use full clone for versioning

wkentaro:
Fixes #17.

The publish workflow triggered on a `v*` tag with **no test gate** and a **shallow checkout** — so a tag with red tests could ship, and `hatch-vcs` fell back to `0.0.0.dev0` instead of deriving the version from the tag.

## Summary
- Make `test.yml` reusable via `workflow_call`, and add a `test` job to `publish.yml` that calls it; `publish` is `needs: test`, so publishing cannot run unless the full matrix passes (no test-matrix duplication).
- Check out the publish job with `fetch-depth: 0` so `hatch-vcs` sees the tags and derives the real version.
- Add a step that lists `dist/` and fails (before upload) if the built wheel does not match the tag version.

## Verification
- [x] Gating: `publish` `needs: test` → a red cell blocks publish (standard GHA semantics).
- [x] `fetch-depth: 0` on the publish checkout.
- [x] Simulated a tag build in a full clone at `v0.2.0` → version `0.2.0`, wheel `git_hunk-0.2.0-py3-none-any.whl`, and the verify step passes.
- [x] `make lint` (yamlfix) clean.

## Out of scope (follow-up: #37)
Further supply-chain hardening — SHA-pinning actions, `environment: pypi` protection, splitting build/publish jobs — is tracked in #37 (some needs repo-settings config; pinning needs the real current SHAs, not guesses).

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -4,13 +4,28 @@ on:
   push:
     tags: [v*]
 jobs:
+  test:
+    uses: ./.github/workflows/test.yml
   publish:
+    needs: test
     runs-on: ubuntu-latest
     permissions:
       contents: read
       id-token: write
     steps:
       - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
       - uses: astral-sh/setup-uv@v6
       - run: uv build
+      - name: Verify built version matches tag
+        run: |
+          version="${GITHUB_REF_NAME#v}"
+          ls -1 dist/
+          test -f "dist/git_hunk-${version}-py3-none-any.whl" || {
+            echo "::error::no wheel for tag ${GITHUB_REF_NAME} (expected version" \
+              "${version}); check fetch-depth and that the tag is a canonical" \
+              "PEP 440 version"
+            exit 1
+          }
       - uses: pypa/gh-action-pypi-publish@release/v1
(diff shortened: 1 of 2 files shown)


wkentaro:

`publish` now declares `needs: test`, so the release job cannot start unless the full 3-OS x 4-Python matrix completes green. `fetch-depth: 0` on the publish checkout is exactly what `hatch-vcs` (confirmed `[tool.hatch.version] source = "vcs"`) needs to walk tag history and derive the real version, fixing the `0.0.0.dev0` regression. Reusing `test.yml` via `workflow_call` avoids duplicating the matrix, the added `test -f dist/...whl` step is a useful pre-upload sanity check, and the trigger stays `push: tags: [v*]`. No YAML or job-graph issues. No blocking issues.

PR #42, opened 2026-06-10
fix: make the skills loader robust to malformed SKILL.md

### LINKED ISSUE #20: Make skills loader robust to malformed or unreadable SKILL.md
## What to build

Make skill loading robust to malformed or unreadable `SKILL.md` files instead of
crashing with an uncaught traceback. Today an unreadable file (permissions) or
non-UTF-8 content raises during read, and frontmatter with a missing closing
`---` is parsed as if the entire file body were frontmatter.

Skip or report bad skill files gracefully, and require a properly terminated
frontmatter block.

## Acceptance criteria

- [ ] An unreadable or non-UTF-8 `SKILL.md` does not crash `skills list`/`get`;
      it is skipped or reported with a clear message.
- [ ] Frontmatter without a closing `---` is treated as having no valid
      frontmatter (keys from the body are not absorbed).
- [ ] Unit tests cover the missing-closing-delimiter case and an unreadable file.

## Blocked by

None - can start immediately.

wkentaro:
Fixes #20.

The skills loader crashed with an uncaught traceback when a `SKILL.md` was unreadable (permissions) or non-UTF-8, and frontmatter with no closing `---` was parsed as if the entire file body were metadata (body keys leaked into `name`/`description`).

## Summary
- `load_skills` skips a `SKILL.md` it cannot read or UTF-8-decode (`OSError`/`UnicodeDecodeError`), printing a one-line warning to stderr and continuing instead of crashing. (Uses plain `sys.stderr` — `_ui` imports `_skills`, so routing through the Rich console would be a circular import.)
- `_parse_frontmatter` requires a terminating `---`; without it, it returns `{}` so body lines are never absorbed as keys (the skill still loads, falling back to its directory name).

## Test plan
- [x] unit: missing-closing-delimiter returns `{}`; an unclosed-frontmatter skill loads with the directory name; a non-UTF-8 file is skipped with a warning; an unreadable (chmod 000) file is skipped (skipif Windows/root): `tests/unit/_skills/`
- [x] verified end-to-end: `GIT_HUNK_SKILLS_DIR` pointing at a non-UTF-8 `SKILL.md` makes `git-hunk skills list` warn and print `No skills.` instead of a traceback
- [x] `make lint` and `uv run pytest` (99 passed)

### DIFF
--- a/git_hunk/_skills.py
+++ b/git_hunk/_skills.py
@@ -6,6 +6,7 @@ the content an agent loads always matches the installed git-hunk version.
 
 import os
 import re
+import sys
 from dataclasses import dataclass
 from pathlib import Path
 
@@ -34,7 +35,11 @@ def load_skills() -> list[Skill]:
         skill_md = child / "SKILL.md"
         if not skill_md.is_file():
             continue
-        content = skill_md.read_text(encoding="utf-8")
+        try:
+            content = skill_md.read_text(encoding="utf-8")
+        except (OSError, UnicodeDecodeError) as exc:
+            print(f"warning: could not read skill {skill_md}: {exc}", file=sys.stderr)
+            continue
         meta = _parse_frontmatter(content)
         skills.append(
             Skill(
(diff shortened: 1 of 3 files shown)


wkentaro:

`load_skills` now catches `OSError`/`UnicodeDecodeError` and skips the offending file with a one-line stderr warning instead of crashing, and `_parse_frontmatter` requires a closing `---` before returning any keys, preventing body content from leaking into metadata. The direct `sys.stderr` use (rather than the Rich console used elsewhere) is justified: `_ui` imports `_skills`, so routing the warning back through `_ui` would be a circular import. All three new behaviors (non-UTF-8 file, unclosed frontmatter, unreadable file) have dedicated unit tests with meaningful assertions. No blocking issues.

PR #41, opened 2026-06-10
fix: harden binary and special-file handling

wkentaro:
Fixes #19.

Hardens handling of binary and special-status files and adds the missing e2e coverage for the binary staging path.

## Summary
- **`list` distinguishes binary state**: `Binary file (modified|added|deleted)` instead of one identical `Binary file` header (derived from the `deleted file mode` / `new file mode` extended-diff headers).
- **Mode-only changes surfaced**: a `chmod` that produces no `@@` hunk now appears as a `Mode <old> -> <new>` hunk instead of being silently dropped, and can be staged/unstaged/discarded.
- **Clean `show`**: binary/mode hunks print their header and return, fixing the spurious blank `  1 ` numbered line from splitting an empty diff body.
- **`-l` rejected** on binary/mode hunks with a clear message (was a misleading "hunk has 0 lines").
- Routing keys on the empty diff (`_is_whole_file_hunk`) so binary and mode-only changes share one whole-file `git add` path; updated SKILL.md.

## Test plan
- [x] unit: parse of modified/added/deleted binary headers and mode-only change: `tests/unit/_hunk/parse_diff_test.py`
- [x] e2e: stage/unstage/discard a modified binary; stage a deleted binary; added-binary header + unstage; mode-only stage/discard (skipped on Windows where git ignores unix modes); clean `show`; `-l` rejection: `tests/e2e/binary_test.py`
- [x] `make lint` and `uv run pytest` (105 passed)

## Follow-up: #40
The change-kind/mode is carried in the free-text `header` string; modeling it as structured `Hunk` fields (and exposing it in `--json`) is tracked in #40, coordinated with the JSON-schema work in #23/#27.

PR #34, opened 2026-06-10
fix: escape user-controlled text in Rich console output

wkentaro:
Fixes #14.

File paths and error text were interpolated straight into Rich markup strings, so a path like `src/[id].tsx` had its `[id]` token swallowed as an unknown markup tag and dropped. The same hazard applied to error/tip messages carrying user-supplied hunk IDs, skill names, and subcommand names.

## Summary
- Wrap every user-controlled value rendered through a Rich markup string with `rich.markup.escape`: tracked and untracked file listings, the `show` file header, and `print_error` msg/tip.
- Trusted markup constants (the `USAGE_*` strings, section headers) are deliberately left unescaped. `Text.append`/`Text(...)` sites (hunk diff bodies, the applied message) were already markup-safe and are untouched.

## Test plan
- [x] e2e: `list` and `show` render `src/[id].tsx` verbatim (tracked and untracked), and an error referencing `[bogus]` prints it literally: `tests/e2e/markup_test.py`
- [x] unit: `print_error` escapes brackets in both message and tip: `tests/unit/_ui/print_error_test.py`
- [x] `make lint` and `uv run pytest` (101 passed)

PR #43, opened 2026-06-10
feat: accept a file path as shorthand for all hunks in a file

### LINKED ISSUE #21: Accept a file path as shorthand for all hunks in a file
## What to build

Let `stage` (and `unstage`/`discard`) accept a file path as shorthand for "all
hunks in that file", so agents and users don't have to enumerate every hunk ID.
Today only hunk IDs are accepted; staging everything in a large file means
collecting and passing dozens of IDs (real usage has hit 51 in one command).

When an argument resolves to a tracked file with changes, operate on all of its
hunks; otherwise treat it as a hunk ID (preserve prefix matching). Define and
document the disambiguation rule when an argument could be both.

## Acceptance criteria

- [ ] `git-hunk stage path/to/file.py` stages every hunk in that file.
- [ ] The same path shorthand works for `unstage` and `discard`.
- [ ] Disambiguation between a path and a hunk ID is deterministic and
      documented; an ambiguous argument produces a clear error or documented
      precedence.
- [ ] e2e tests cover staging a whole file by path and the disambiguation rule.

## Blocked by

None - can start immediately.

wkentaro:
Fixes #21.

`stage`, `unstage`, and `discard` now accept a **file path** as shorthand for every hunk in that file, so agents and users don't have to enumerate hunk IDs (real usage has hit 51 IDs in one command).

## Summary
- New `_select_hunks` resolves each argument: an argument that matches a changed file's path selects all of that file's hunks; otherwise it's a hunk ID (prefix match). Results are deduped by hunk ID, order preserved.
- **Disambiguation (documented in SKILL.md + `--help`):** a path takes precedence when an argument could be both. Because hunk IDs are lowercase hex, a non-hex argument can only be a path — so a typo like `stage src/typo.py` reports `no changed file matches 'src/typo.py'` instead of a misleading "hunk not found". Arguments are normalized (`./src/foo.py` matches `src/foo.py`).

## Test plan
- [x] e2e: stage/unstage/discard a whole file by path; mixing path + ID; dedup (path + an ID it contains); path precedence over an ID-lookalike filename; `./`-prefixed path; `-l` rejected on a multi-hunk path; clear errors for a missing path vs a missing hunk ID: `tests/e2e/path_shorthand_test.py`
- [x] dogfooded: this PR's own change was staged with `git-hunk stage <paths>`

## Scope notes
Kept the documented path-precedence rule (the issue allows "error OR documented precedence") rather than adding a `git`-style `--` escape hatch — the only "ambiguous" case is a changed file named exactly like a current hex hunk ID, which is vanishingly rare and still stageable via the file. `show` remains ID-only (out of scope).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ and this project adheres to
 - `skills` subcommand (`git-hunk skills list|get|path`) serving the bundled,
   version-matched core usage guide for AI agents.
 - Examples section in `--help` for every subcommand.
+- Accept a file path as shorthand for all hunks in a file, so
+  `git-hunk stage src/foo.py` stages every hunk in that file (#21).
 
 ### Changed
 
--- a/git_hunk/_cli.py
+++ b/git_hunk/_cli.py
@@ -1,4 +1,6 @@
 import json
+import os
+import re
 from dataclasses import replace
 
 import click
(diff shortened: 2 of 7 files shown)


wkentaro:

`_select_hunks` dispatches path-first then hex-ID with dedup and order preservation; `os.path.normpath` handles `./`-prefixes, and SHA256 hex is always lowercase so `[0-9a-f]+` matches every valid hunk ID and nothing else. The error-test change from `"nonexistent"` to `"deadbee"` is a necessary adaptation to the new routing (a non-hex token now routes to the path-missing branch). Coverage is thorough: stage/unstage/discard by path, `./`-prefix, mixed path+ID, dedup, hex-lookalike precedence, `-l` rejection on a multi-hunk path, and a clear missing-path error. No blocking issues; CI green.

wkentaro:
resolve test failure

wkentaro:

Resolved. The failure appeared after rebasing onto the current `main`, where this PR's path-shorthand routing collided with the now-merged #34 (Rich markup escaping) and #35 (empty-id validation): a non-hex argument like `""` or `[bogus]` was being sent straight to file-path lookup, which bypassed #35's empty-id guard and changed the `[bogus]` error message.

Two fixes, folded into the commit:
1. Re-added an empty/whitespace guard at the top of `_select_hunks`, so `discard ""` still fails with "must not be empty" (preserves #35).
2. Updated #34's `test_error_renders_bracketed_value_verbatim` to the new path-shorthand message (`no changed file matches '[bogus]'`) — it still asserts the brackets render verbatim, so #34's escaping guarantee is intact.

Rebased onto current `main`; full local suite and CI are green (13/13).

wkentaro:
need rebase

wkentaro:

Already rebased — `feat/path-shorthand` is on current `main` (`81164d1`, behind=0) after resolving the `_cli.py` and `_ui.py` conflicts against the merged #33 (`reverse` threading) and #41 (`whole_file` handling). `mergeStateStatus` is CLEAN and CI is green (13/13).

wkentaro:
rebase and update changelog

wkentaro:

Done. Rebased onto current `main` (`5ab69f4`) and added an `### Added` entry to `CHANGELOG.md`:

> - Accept a file path as shorthand for all hunks in a file, so `git-hunk stage src/foo.py` stages every hunk in that file (#21).

Folded into the feature commit; `mdformat --check` passes. CI re-running.

PR #45, opened 2026-06-13
feat: add --dry-run to stage, unstage, and discard

wkentaro:
## Summary
- Add a `--dry-run` flag to `stage`, `unstage`, and `discard` that reports what would change without modifying the index or working tree.
- Text hunks are validated with `git apply --check`, so the exit code reflects whether the operation would actually have succeeded.
- Whole-file (binary / mode-only) selections are reported but left untouched.
- Output uses a `would stage / unstage / discard` verb to distinguish a preview from a real apply.

A safety net before the destructive `discard` in particular, and it lets agents preview an operation before committing to it.

## Test plan
- [x] tests added: `tests/e2e/dry_run_test.py` (stage/unstage/discard change nothing, `-l` preview, binary preview, unknown id exits non-zero)
- [x] `uv run pytest` (164 passed)
- [x] `uv run ruff check` / `ruff format --check` / `uv run ty check`

Closes #25

PR #46, opened 2026-06-13
feat: add a commit subcommand to stage and commit in one step

### LINKED ISSUE #26: Add a commit subcommand (stage + commit in one step)
## What to build

Add a `commit` subcommand that stages the given hunks and commits them in one
step, collapsing the common `git-hunk stage <ids> && git commit -m "..."` pattern
into a single call. Real usage shows a meaningful share of genuine two-step
flows this would shorten; keep the existing stage/commit split available for
workflows that verify with `git diff --cached` in between.

## Acceptance criteria

- [ ] `git-hunk commit <ids> -m "msg"` stages exactly those hunks and creates one
      commit with that message.
- [ ] Supports the same hunk-ID/prefix matching (and `-l`) semantics as `stage`.
- [ ] Behavior when nothing matches, or the commit fails, is clear and leaves a
      predictable index state.
- [ ] e2e tests cover a successful stage+commit and the no-match/failure paths.

## Blocked by

None - can start immediately.

wkentaro:
## Summary
- Add a `commit` subcommand: `git-hunk commit <ids|files> -m "msg"` stages the selected hunks and commits them in one step, collapsing the common `git-hunk stage <ids> && git commit -m "..."` flow.
- Same selection semantics as `stage`: id prefixes, file-path shorthand, and `-l` line selection.
- Aborts if anything is already staged, so the commit holds exactly the selected hunks (no silent inclusion of unrelated pre-staged changes).
- If the commit is rejected (e.g. by a pre-commit hook), the hunks are left staged so you can retry with `git commit`.
- The staging core is factored out of `_run_patch_command` into a shared `_apply_selection`, so stage/unstage/discard and commit share one implementation.

## Test plan
- [x] tests added: `tests/e2e/commit_test.py` (single/multiple/by-file/partial-line commits, message required, unknown id, dirty-index abort, pre-commit-hook failure leaves hunks staged)
- [x] `uv run pytest` (181 passed)

Closes #26

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,6 +105,16 @@ git-hunk discard d161935               # restore from HEAD
 git-hunk discard d161935 -l ^3,^5-7    # discard excluding specific lines
 ```
 
+### Commit
+
+```bash
+git-hunk commit d161935 -m "fix: ..."      # stage a hunk and commit it in one step
+git-hunk commit d161935 -l 3,5-7 -m "..."  # stage specific lines and commit
+```
+
+`commit` aborts if anything is already staged, so the commit contains exactly
+the selected hunks.
+
 ### JSON output
 
 ```bash
--- a/git_hunk/_cli.py
+++ b/git_hunk/_cli.py
@@ -7,6 +7,7 @@ import click
 
 from . import __version__
 from ._git import apply_patch
+from ._git import commit
 from ._git import discard_files
 from ._git import get_diff
 from ._git import get_untracked_files
@@ -22,6 +23,7 @@ from ._skills import Skill
 from ._skills import load_skills
 from ._skills import skills_root
 from ._ui import HELP
+from ._ui import HELP_COMMIT
 from ._ui import HELP_DISCARD
 from ._ui import HELP_LIST
 from ._ui import HELP_SHOW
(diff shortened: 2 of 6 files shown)


wkentaro:

The `commit` subcommand is correct on every critical path: the dirty-index guard (`get_diff(staged=True)`) fires **before** any staging, so an already-staged index aborts with nothing touched; unknown-id and empty-message errors abort before mutating the index; and a rejected commit (e.g. a pre-commit hook) is caught, re-raised as a non-zero `CliError`, and intentionally leaves the hunks staged (documented and tested). The refactor extracting a shared `_apply_selection` from `_run_patch_command` avoids duplicating any staging logic. Tests cover the happy path, dirty-index abort, hook rejection, and id/file/`-l` selection, asserting actual commit and index contents rather than just exit codes.

Also fixed the failing `lint` job — `SKILL.md` needed an `mdformat` reflow — folded into the commit; full lint + suite (181) pass locally and CI is re-running.

Non-blocking: no `--dry-run` on `commit` yet (a reasonable omission, and a natural follow-up now that #45 landed), and the test's module-level helpers could be fixtures.

PR #47, opened 2026-06-13
refactor: remove unreachable auto-split logic

### LINKED ISSUE #24: Resolve dead auto-split logic (remove or fix and cover)
## What to build

Resolve the dead auto-split logic. The hunk auto-splitting code requires a gap
of 7 context lines between change regions, but git's default `-U3` never
produces a single hunk with a gap that large (it splits at >6), so the split
path is effectively unreachable in normal use. A related defect: when the split
path IS reached (non-default context), sub-hunks after the first reuse the first
hunk's function-context tail in their `@@` header.

Decide whether to remove the unused split machinery (simplify) or make it
correct and reachable, then either delete it or fix the tail-inheritance bug and
add tests that actually exercise it.

## Acceptance criteria

- [ ] Either: the dead split code is removed and the module is simpler with all
      tests still passing; or: splitting is made reachable/correct, the sub-hunk
      header tail bug is fixed, and tests exercise a real multi-region split.
- [ ] No behavior change for the default `-U3` workflow.

## Blocked by

None - can start immediately.

wkentaro:
## Summary
- Remove the hunk auto-split machinery from `_hunk.py` (`_split_hunk`, `_build_sub_hunks`, `_find_change_regions`, `_find_split_points`, `_advance_offsets`, `_SubHunk`) and make `parse_diff` emit one hunk per `@@` section. `_hunk.py` drops from 300 to ~170 lines.
- The splitter was dead code: git-hunk runs `git diff` at `-U3`, where git already separates changes >6 context lines apart into distinct `@@` sections, so a single section with the `>=7`-line gap the splitter required can never occur. It also carried a latent bug (later sub-hunk `@@` headers inherited the first sub-hunk's function-context tail).
- Pin `-U3` explicitly in `get_diff` so the "one `@@` section per hunk" assumption is no longer left to git's default.

## Why remove rather than fix-and-reach?
Making the splitter reachable would mean splitting finer than git's own `-U3` hunking — but `-l` line selection already provides finer-than-hunk granularity, so the splitter is redundant as well as unreachable.

## Behavior
No change for real `-U3` diffs (verified: N distant changes → N `@@` sections → N hunks, before and after). I confirmed empirically that neither `diff.unified` config nor `GIT_DIFF_OPTS` alters git's hunk boundaries in practice (only an explicit `-U` flag does, which git-hunk never passes); the `-U3` pin makes the assumption explicit regardless.

## Test plan
- [x] Removed `split_test.py` and the synthetic split-path tests (they exercised input `git diff -U3` cannot produce); kept/renamed `test_one_section_maps_to_one_hunk`
- [x] `uv run pytest` (163 passed)

Closes #24


wkentaro:

The dead-code claim holds rigorously. The old splitter required a gap of >=7 context lines *within a single `@@` section* (`MIN_GAP = 2*3 + 1 = 7`), but `git diff -U3` merges change regions within 6 context lines into one section and emits separate `@@` sections for anything further apart — so the maximum intra-section gap is 6, one short of the trigger. The splitter could never fire. This PR also pins `-U3` explicitly in `get_diff`, locking the invariant instead of relying on git's default, and the e2e multi-hunk tests confirm distant changes still yield multiple hunks via git's own `@@` sectioning. No removed helper is referenced anywhere on the branch, the `TypedDict` import is cleaned up, and the deleted `split_test.py` / split-path cases covered only the now-removed code (no-newline marker coverage is preserved in `test_preserves_no_newline_marker_in_diff` and `test_marker_not_counted_as_change`). CI green (13/13).

Issue #44, opened 2026-06-13
Emit byte-safe JSON for non-UTF-8 paths and content (lone surrogates break strict parsers)


wkentaro:
## Context

Surfaced during review of #11 / PR #32 (decode git output with `surrogateescape`). That fix is correct and tested; this is a separate downstream gap in the `--json` output, not a regression in #32.

## Problem

With `surrogateescape` decoding, a non-UTF-8 byte in a path or diff line becomes a lone surrogate in the Python `str` (e.g. `0xe9` -> `\udce9`). The `--json` commands serialize these with `json.dumps(..., ensure_ascii=True)` (the default), which emits the lone surrogate as a `\udcXX` escape:

```
$ git-hunk list --json
[{"file": "pass\udce9.txt", ...}]
```

This is a lone-surrogate JSON string, which downstream consumers handle inconsistently:

- Python's own `json.loads` round-trips it.
- `jq` (1.8.x) parses it but **replaces it with U+FFFD** — the original byte is silently lost (lossy).
- Strict parsers (Go `encoding/json`, Rust `serde_json`, older `jq`) **reject** the document outright.

So `list --json` on a repo containing a non-UTF-8 filename or diff content either corrupts the value or breaks the consumer. Affected sinks: the three `json.dumps(...)` calls in `git_hunk/_cli.py` (`list --json`, and the per-command JSON outputs).

## Proposal

Pick a byte-safe representation for non-UTF-8 text in JSON output (maintainer's call):

1. Replace undecodable bytes with U+FFFD before serializing (valid, but lossy).
2. Emit the raw bytes in a separate, explicitly-encoded field (e.g. base64 / percent-escaped), keeping the human field best-effort.
3. Document that `--json` assumes UTF-8 paths and fail loudly with a clear error otherwise.

## Acceptance criteria

- [ ] `list --json` on a repo with a non-UTF-8 path or diff line produces output that strict JSON parsers (`jq`, Go, Rust) accept.
- [ ] The chosen representation is covered by a test that asserts the output parses cleanly.

## Scope notes

- Distinct from #23 (document and version the `--json` schema): that is about the schema's shape and versioning; this is about byte-safety of the encoding. The two can be addressed together but are not the same gap.
- Distinct from #11 / #32 (the decode fix itself), which is already merged/correct.

wkentaro:
> *This was generated by AI during triage.*

## Agent Brief

**Category:** bug
[...]

PR #48, opened 2026-06-13
feat: include context_before in --json output

### LINKED ISSUE #27: Include context_before in --json output (parity with human display)
## What to build

`Hunk.to_dict()` (`git_hunk/_hunk.py:21-30`) builds the per-hunk object emitted
by `--json`, but it omits the `context_before` field. That field is computed and
stored on every `Hunk` (`_extract_context_before`, the function-name annotation
from the `@@ ... @@ <context>` header) and is shown in the human-readable `list`
display (`git_hunk/_ui.py:44-45` and `157-158`), yet `--json` consumers never
see it.

git-hunk targets AI agents as a primary consumer, and those agents read the
`--json` output rather than the rich display. Dropping `context_before` means an
agent cannot tell which function/section a hunk belongs to without re-parsing the
`@@` header itself.

## Proposed change

Add `"context_before": self.context_before` to the dict returned by `to_dict()`,
giving JSON output parity with the human display.

## Acceptance criteria

- [ ] `git-hunk list --json` includes `context_before` for each hunk, matching
      what the human `list` display shows.
- [ ] Existing tests still pass; a test asserts `context_before` is present in
      the serialized dict.
- [ ] No change to the human-readable output.

## Related

- #23 (Document and version the --json output schema) — this adds a field to the
  public `--json` format, so sequence it with the schema-versioning work to avoid
  changing the documented schema twice.

wkentaro:
## Summary
- Add `context_before` to `Hunk.to_dict()` so `git-hunk list --json` carries the function/section annotation (the text git puts after the `@@` header) that the human `list` display already shows.
- git-hunk targets AI agents, which read `--json` rather than the rich display; without this they couldn't tell which function a hunk belonged to without re-parsing the header.
- The field is always present and is an empty string when git provides no context (plain text / binary / mode-only hunks). README `--json` example and a note updated to match.

## Test plan
- [x] unit: `to_dict` includes `context_before` (named-context and empty-string cases)
- [x] e2e: `list --json` exposes `context_before` matching the function context shown in the display
- [x] `uv run pytest` (176 passed) · `ruff check` / `ruff format --check` / `ty check`

## Note for the maintainer (out of scope here)
While testing I noticed the **human `list`/applied display shows the function context twice** for hunks that have it — once inside `header` (git emits `@@ … @@ def foo():`) and once via the separate `context_before` column, e.g. `@@ -2,7 +2,7 @@ def foo():  def foo():`. This PR deliberately does not touch it, since #27's acceptance criteria require no change to the human-readable output. Happy to file a follow-up to render the bare `@@` header (or drop the duplicate column) if you'd like.

Relatedly, #23 (version the `--json` schema) will document this field as part of the versioned contract; this change is additive and safe for current consumers.

Closes #27

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,13 +127,17 @@ git-hunk list --json
     "id": "d161935",
     "file": "src/main.py",
     "status": "unstaged",
-    "header": "@@ -10,3 +10,5 @@",
+    "header": "@@ -10,3 +10,5 @@ def main():",
+    "context_before": "def main():",
     "additions": 2,
     "deletions": 0,
     "diff": "..."
   }
 ]
 ```
+
+`context_before` is the function/section git names after the `@@` header; it is
+an empty string when git provides no context (e.g. plain text or binary hunks).
 
 ## Comparison
 
--- a/git_hunk/_hunk.py
+++ b/git_hunk/_hunk.py
@@ -30,6 +30,7 @@ class Hunk:
             "file": self.file,
             "status": self.status,
             "header": self.header,
+            "context_before": self.context_before,
             "additions": self.additions,
             "deletions": self.deletions,
             "diff": self.diff,
--- a/tests/e2e/list_test.py
+++ b/tests/e2e/list_test.py
@@ -151,3 +151,19 @@ def test_list_status_field_in_json(cli: GitHunkCLI) -> None:
     hunks = cli.run_json("list", "--json")
     assert all("status" in h for h in hunks)
     assert hunks[0]["status"] == "unstaged"
+
+
+def test_list_context_before_field_in_json_matches_display(cli: GitHunkCLI) -> None:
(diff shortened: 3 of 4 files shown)


wkentaro:

`context_before` was already a first-class `Hunk` field (populated by `_extract_context_before` during parsing); this PR simply surfaces it in `to_dict()` so `list --json` carries the same function/section annotation the human `list` display already shows. The change is purely additive — no impact on human output or any other serialization — and both the populated and empty-string cases are covered by unit tests, with an e2e test asserting `--json`/display parity. The README documents the field accurately. CI green (13/13).

Non-blocking: as more fields accrete on the `--json` contract, the `schema_version` work in #23 (and the byte-safety follow-up #44) becomes more valuable.

PR #49, opened 2026-06-13
feat!: version the --json output schema

### LINKED ISSUE #23: Document and version the --json output schema

wkentaro:
## Summary
- Wrap `list --json` in a versioned envelope, `{"schema_version": 1, "hunks": [...]}`, so downstream agents/tools can depend on a documented, versioned shape instead of guessing at a bare array.
- Add a `JSON_SCHEMA_VERSION` constant (single source of truth, imported by the contract test).
- Document the full schema in the README (envelope + field table with types and `status` values) and note the compatibility rule (additive = non-breaking; rename/remove/retype = version bump).
- Add a contract test that pins the envelope and the required hunk keys (subset check, so it tolerates additive fields) and fails on any breaking shape change.
- Tests read hunks through a `run_list_json` helper that unwraps the envelope; ~57 call sites migrated.

**Breaking:** the bare-array output is gone (noted in CHANGELOG).

## Test plan
- [x] contract test: envelope shape + `schema_version` + required hunk keys (`tests/e2e/list_test.py`), incl. the empty case
- [x] `uv run pytest` (173 passed) · `ruff check` / `ruff format --check` / `ty check`
- [x] manual: `list --json` emits the envelope for both empty and non-empty trees

## Notes for the maintainer
- The envelope shape (`schema_version` + `hunks`) is a one-way-door public-API decision; flagging it explicitly in case you'd prefer a different key/layout — easy to adjust.
- Scoped to `list --json` per the acceptance criteria; `skills --json` keeps its own (simpler, bare-array) shape.
- Sequences cleanly with #27 (PR #48, adds `context_before` to each hunk): the contract test uses a required-keys *subset* check, so that additive field won't break it, and adding it won't need a `schema_version` bump.

Closes #23

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,9 @@ and this project adheres to
 
 ### Changed
 
+- **Breaking:** `list --json` now wraps its output in a versioned envelope,
+  `{"schema_version": 1, "hunks": [...]}`, instead of a bare array, so consumers
+  can depend on a documented, versioned shape (#23).
 - README image paths are rewritten to absolute URLs so they render on PyPI.
 
 ### Fixed
--- a/README.md
+++ b/README.md
@@ -121,23 +121,43 @@ the selected hunks.
 git-hunk list --json
 ```
 
+`list --json` emits a versioned envelope so consumers can depend on a stable
+shape:
+
 ```json
-[
-  {
-    "id": "d161935",
-    "file": "src/main.py",
-    "status": "unstaged",
-    "header": "@@ -10,3 +10,5 @@ def main():",
-    "context_before": "def main():",
-    "additions": 2,
-    "deletions": 0,
-    "diff": "..."
-  }
-]
+{
+  "schema_version": 1,
+  "hunks": [
+    {
+      "id": "d161935",
+      "file": "src/main.py",
+      "status": "unstaged",
+      "header": "@@ -10,3 +10,5 @@ def main():",
+      "context_before": "def main():",
+      "additions": 2,
+      "deletions": 0,
+      "diff": "..."
+    }
+  ]
+}
 ```
 
-`context_before` is the function/section git names after the `@@` header; it is
-an empty string when git provides no context (e.g. plain text or binary hunks).
+| Field            | Type   | Description                                                                                        |
+| ---------------- | ------ | -------------------------------------------------------------------------------------------------- |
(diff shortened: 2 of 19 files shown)


wkentaro:

The versioned envelope `{"schema_version": 1, "hunks": [...]}` is emitted correctly, and `JSON_SCHEMA_VERSION` is a single constant imported by the contract test (never hardcoded twice). The migration is complete — all 15 e2e files now read `--json` through the new `run_list_json` helper with no stale bare-array sites — and the contract test uses a subset check (`_REQUIRED_HUNK_KEYS <= hunk.keys()`), so additive fields like #48's `context_before` won't break it. The breaking change (bare array → envelope) is intentional for establishing the #23 contract and is documented in CHANGELOG/README/SKILL.md; scoping the envelope to `list --json` while leaving `skills --json` a bare array is the correct scope.

Fixed the failing `lint` job (mdformat reflow of `README.md`), folded into the commit; full lint + suite (173) pass locally and CI is re-running.

Merge-ordering note: this and #48 (`context_before`) both touch `--json`, so whichever merges second needs a trivial rebase — the contract test already tolerates the extra field.

wkentaro:

Rebased onto current `main` after #48 merged. Resolved the `README.md` conflict by keeping this PR's versioned-envelope docs and folding in #48's `context_before` field (added to both the JSON example and the schema field table). Also migrated #48's new `test_list_context_before_field_in_json_matches_display` from `run_json` to the envelope-aware `run_list_json` helper — consistent with this PR's migration; without it that test read the old bare array and failed against the envelope.

Full suite (176) + lint pass locally; CI re-running. Will re-apply `recommend-merge` once it's green.

PR #51, opened 2026-06-13
fix: keep full context_before when section heading contains @@

wkentaro:
The hunk-header parser pulls the section heading after the `@@ … @@` range with
a greedy regex (`@@.*@@`). When that heading itself contains `@@` — a Ruby class
variable like `@@count`, or a trailing `# @@` comment — the match ran to the
*last* `@@` on the line and truncated the heading, so `context_before` (shown in
the human listing and emitted in `--json`) lost everything up to the embedded
`@@`.

Anchoring to the first closing `@@` pair with a lazy quantifier (`@@.*?@@`) fixes
it: the range between the delimiters never contains `@@`, so the first pair is
always the structural one. Also drops a redundant double `strip()` in the same
return.

## Test plan
- [x] tests added: `tests/unit/_hunk/extract_context_before_test.py` — `test_anchors_to_first_at_at_pair` fails on the old greedy regex, passes on the fix
- [x] `make test` (182 passed) and `make lint` clean

### DIFF
--- a/git_hunk/_hunk.py
+++ b/git_hunk/_hunk.py
@@ -83,8 +83,8 @@ def extract_file_path(file_diff: str) -> str | None:
 
 
 def _extract_context_before(header: str) -> str:
-    match = re.search(r"@@.*@@\s*(.*)", header)
-    return match.group(1).strip() if match and match.group(1).strip() else ""
+    match = re.search(r"@@.*?@@\s*(.*)", header)
+    return match.group(1).strip() if match else ""
 
 
 def _whole_file_hunk(filepath: str, header: str) -> Hunk:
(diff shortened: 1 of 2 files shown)


wkentaro:

Correct, minimal fix. The unified-diff hunk-header range (`-a,b +c,d`) contains only digits, commas, spaces, and `+`/`-`, so it can never contain `@@`; the lazy `@@.*?@@` therefore always locks onto the structural delimiter pair, and for a plain heading it produces output identical to the old greedy regex. The redundant double `strip()` removal is semantically equivalent across all inputs (empty, whitespace-only, normal, leading/trailing spaces). The regression test `test_anchors_to_first_at_at_pair` provably fails on the old greedy code (it would return `'tag'` instead of `'class Foo:  # @@ tag'`) and passes on the fix, with surrounding cases covering plain, empty, and whitespace-only headings. CI green (13/13).

PR #59, opened 2026-06-18
chore: pipe git ls-files into xargs in Makefile

wkentaro:
Adopt labelme's xargs style in the `lint` and `format` targets: pipe `git ls-files` into `xargs` instead of expanding it inline with `$(shell ...)`. This keeps the file-globbed tool invocations consistent with labelme's Makefile and makes the printed command line show the pipeline rather than the full expanded file list.

## Test plan
- [x] `make lint`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,17 +22,17 @@ lint:  # Lint code
 	$(call exec,uv run ruff format --check)
 	$(call exec,uv run ruff check)
 	$(call exec,uv run ty check --no-progress)
-	$(call exec,uv run taplo fmt --check $(shell git ls-files "*.toml"))
-	$(call exec,uv run mdformat --check $(shell git ls-files "*.md"))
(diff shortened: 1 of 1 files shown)


wkentaro:

**Verdict: \`recommend-merge\`**

Single-commit chore converting `$(shell git ls-files ...)` argument expansion into `git ls-files ... | xargs` pipelines for `taplo`/`mdformat`/`yamlfix` in the `lint` and `format` targets.

- Touches only `Makefile`; one clean, correctly-scoped commit.
- CI is fully green, and the `lint` job runs these exact commands, so the new pipeline form is exercised against the real file set.
- No rebase needed: `main` gained 0 commits since the merge-base and there is no file overlap.

Minor (non-blocking): `xargs` here has no `-r`/`--no-run-if-empty`, so if any glob ever matched zero tracked files the formatter would be invoked with no arguments. Harmless today since `*.toml`, `*.md`, and `*.yml`/`*.yaml` all exist in the repo; worth `-r` only if that stops being true.

PR #60, opened 2026-06-18
fix: partial line-staging corrupts no-newline-at-EOF files

### LINKED ISSUE #54: Partial line-staging corrupts no-newline-at-EOF files (last line merged with addition)
## Bug

Partial line-staging (`stage <id> -l <lines>`) of a hunk where the file gains a trailing newline (no-newline -> has-newline) silently **corrupts the file**: the no-newline last line is merged with the staged addition.

Repro:

```console
$ printf 'a\nb' > f.txt          # 'a\nb', NO trailing newline
$ git add f.txt && git commit -m init
$ printf 'a\nB\n' > f.txt         # last line b->B, and gains a trailing newline
$ git-hunk show <id>
  1  a
  2 -b
    \ No newline at end of file
  3 +B
$ git-hunk stage <id> -l 3        # stage ONLY the '+B' addition
$ git cat-file -p :f.txt | xxd
00000000: 610a 6242 0a              a.bB.      # staged index is "a\nbB\n"  -- WRONG
```

Expected, matching how partial-staging a modification behaves for newline-terminated files (`a\nb\n -> a\nB\n` staged at the addition yields `a\nb\nB\n`): the staged index should be `a\nb\nB\n`, keeping `b` and `B` as separate lines.

## Root cause

In `_filter_body_lines` (`git_hunk/_lines.py`), a `\ No newline at end of file` marker is kept whenever the preceding line was kept (`prev_kept`). When an unselected `-b` is converted to **context** and a selected `+B` is appended after it, the marker stays attached to the now-non-final context line:

```
@@ -1,2 +1,3 @@
 a
 b
\ No newline at end of file      <- stale: 'b' is no longer the last line
+B
```

git then applies `+B` onto the no-newline `b`, merging them into `bB`. A no-newline marker is only valid as the **final** line of its side; once added lines follow the line it annotates, the marker must be dropped (and the annotated line gains a trailing newline).

## Affected

- `git_hunk/_lines.py` `_filter_body_lines` (the `prev_kept` marker logic, ~lines 72-75).
- Confirmed direction: no-newline -> has-newline, forward (stage), partial line selection. The reverse (unstage/discard) and both-sides-no-newline variants should be checked for mirror cases.

## Why this needs a human

Silent data corruption. The fix must be proven across both newline directions (no-newline->newline and newline->no-newline), both apply directions (stage/forward and unstage/reverse), and the case where the marker appears on both sides, with round-trip `git apply` e2e tests. A subtly wrong fix trades one corruption for another.

## Related

- #28 (model the no-newline marker as a line property) states the current marker handling is "correct and tested." This bug shows the `prev_kept` flag #28 references is **not** correct for partial staging; the #28 refactor should fix this case, not preserve it.

## Acceptance criteria

- [ ] `stage -l` selecting the addition of a no-newline->newline modification yields separate lines (no merge), verified by an e2e test against real `git apply`.
- [ ] Reverse (unstage/discard) partial selection on the same hunk verified.
- [ ] A both-sides-no-newline hunk verified.
- [ ] Unit tests covering the `_filter_body_lines` marker placement for these cases.

wkentaro:
Fixes #54.

Partial line-staging (`-l`) of a hunk whose last line gains a trailing newline (no-newline -> has-newline) silently corrupted the file: the no-newline last line was merged with the staged addition. `printf 'a\nb'` then staging only the `+B` of `a\nB\n` produced an index of `a\nbB\n` instead of `a\nb\nB\n`.

The cause was in `_filter_body_lines`: a `\ No newline at end of file` marker was kept whenever the preceding line survived. When an unselected `-b` (old EOF, no newline) became context and a selected `+B` was appended after it, the marker stayed glued to the now-non-final context line, so `git apply` merged them.

## Summary
- Split the transform into `_select_body_lines` (folds each marker onto the line it annotates, records the line's original prefix) and `_render_body_lines` (re-places markers).
- Re-place markers by the rule *a no-newline marker is valid only on the final line of the side it describes*. When a converted-context line is no longer its side's final line, split it back into a `-`/`+` pair: the side that truly has no trailing newline keeps the marker, the other side gains a newline.
- The bug case now stages `a\nb\nB\n`, keeping `b` and `B` as separate lines.

## Test plan
- [x] e2e tests for all four acceptance criteria: no-newline->newline stage, unstage, discard, and both-sides-no-newline (`tests/e2e/no_newline_test.py`)
- [x] unit tests asserting exact marker placement on split and on dropped additions (`tests/unit/_lines/filter_test.py`)
- [x] exhaustive round-trip sweep against real `git apply`: every selection subset across 9 scenarios, 127 forward (stage) + 127 reverse (discard), all apply cleanly with no data loss
- [x] `make lint` and full suite (189 passed) green

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,5 +44,8 @@ and this project adheres to
   with a readable error (#15).
 - Constrain the source distribution to the package, tests, and metadata so it
   no longer ships unrelated `tmp/` files (#16).
+- Split the no-newline last line when partial line-staging (`-l`) gives it a
+  trailing newline, so staging only the addition no longer merges it with the
+  added line and corrupts the file (#54).
 
 [unreleased]: https://github.com/wkentaro/git-hunk/compare/v0.2.0...HEAD
--- a/git_hunk/_lines.py
+++ b/git_hunk/_lines.py
@@ -59,7 +59,6 @@ class _BodyLine(NamedTuple):
     prefix: str  # rendered side: ' ' context, '-' old, '+' new
     text: str
     no_newline: bool  # the line it represents has no trailing newline
-    origin: str  # original prefix, before an unselected change became context
 
 
 def _select_body_lines(
@@ -68,11 +67,7 @@ def _select_body_lines(
     *,
     keep_prefix: str,
 ) -> list[_BodyLine]:
-    """Apply the line selection, folding each no-newline marker onto its line.
-
-    ``origin`` records the original prefix so a converted change kept as context
-    can be split back apart later when its no-newline marker has to move.
-    """
+    """Apply the line selection, folding each no-newline marker onto its line."""
     kept: list[_BodyLine] = []
     line_num = 0
     dropped_last = False
@@ -84,9 +79,9 @@ def _select_body_lines(
         line_num += 1
         prefix, text = line[:1], line[1:]
         if prefix == " " or line_num in selected:
-            kept.append(_BodyLine(prefix, text, False, prefix))
-        elif prefix == keep_prefix:
-            kept.append(_BodyLine(" ", text, False, prefix))  # survive as context
+            kept.append(_BodyLine(prefix=prefix, text=text, no_newline=False))
+        elif prefix == keep_prefix:  # unselected change survives as context
+            kept.append(_BodyLine(prefix=" ", text=text, no_newline=False))
         else:
             dropped_last = True
             continue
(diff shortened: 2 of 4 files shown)


wkentaro:

Finalize pass: folded the unpushed `refactor: drop unreachable reverse no-newline split branch` commit and added a unit test (`test_reverse_keeps_no_newline_new_context_without_split`) that pins the reverse + no-newline-context output, making the refactor's "unreachable `+`-origin split branch" claim machine-checkable rather than argument-only.

Validated locally: `make lint` clean, full suite 190 passed (was 189), and a real-CLI smoke of the #54 scenario (stage only `+B`) yields a staged index of `a\nb\nB\n` (three newline-terminated lines), not the corrupted `a\nbB\n`. No rebase needed (main is at the merge-base; zero divergence). Pushed `639405e..88d8012`; awaiting the re-triggered CI before recording a verdict.

wkentaro:

CI resolved green across all 14 checks (lint + test on macOS/Ubuntu/Windows × Python 3.10–3.13) on the finalized head `88d8012`. Merge state is CLEAN against `main` with zero divergence from the merge-base, so no rebase was needed and the green reflects the merged result. The branch was untouched this tick.

Verdict: **recommend-merge**.

PR #62, opened 2026-06-24
refactor: split diff into per-file headers once in build_patch

wkentaro:
`build_patch` called `_get_file_header` once per changed file, and each call
re-split the entire `diff_output` with `re.split(r"(?=^diff --git )")` and scanned
it linearly, making header lookup O(files^2). Replace it with
`_extract_file_headers`, which splits the diff once into a `{path: header}` map
that `build_patch` indexes. Behavior is unchanged; the missing-file `ValueError`
now raises from `build_patch` instead of the helper.

## Test plan
- [x] `make test` (191 passed)
- [x] CLI smoke: `git-hunk stage <id>` partial-stages one of two changed files correctly

### DIFF
--- a/git_hunk/_patch.py
+++ b/git_hunk/_patch.py
@@ -4,25 +4,30 @@ from ._hunk import Hunk
 from ._hunk import extract_file_path
 
 
-def _get_file_header(diff_output: str, filepath: str) -> str:
-    file_diffs = re.split(r"(?=^diff --git )", diff_output, flags=re.MULTILINE)
-    for file_diff in file_diffs:
-        if extract_file_path(file_diff) == filepath:
-            # Return everything up to the first @@ line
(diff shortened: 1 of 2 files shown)


wkentaro:

A behavior-preserving refactor: `build_patch` now extracts every file header once into a dict (`_extract_file_headers`) instead of re-splitting the whole diff per file, turning O(files²) splitting into O(files). The missing-file `ValueError` path is preserved (now raised in `build_patch`), type annotations were added, and `maxsplit=1` is equivalent to the old `parts[0]`.

- A focused correctness review found no actionable issues (the only divergence is duplicate-path handling, which `git diff` never emits).
- No rebase needed: `main` gained 0 commits since the merge-base.
- CI: all 13 checks green (lint + test across Linux/macOS/Windows × Python 3.10–3.13), merge state CLEAN.

PR #63, opened 2026-06-24
refactor: extract split_file_diffs helper for per-file diff splitting

wkentaro:
The `re.split(r"(?=^diff --git )", diff_output, flags=re.MULTILINE)` idiom for
splitting a combined diff into per-file chunks was written verbatim in two
places: `parse_diff` (`_hunk.py`) and `_extract_file_headers` (`_patch.py`).
This extracts it into a `split_file_diffs` helper in `_hunk.py` so the
non-obvious lookahead regex is spelled once and both call sites share it.

Pure refactor, no behavior change. Follows the same dedup pattern as the
recent `strip_trailing_empty_lines` extraction, and matches the existing
non-prefixed cross-module helpers (`extract_file_path`, `strip_trailing_empty_lines`).

## Test plan
- [x] `make test` (191 passed)
- [x] `make lint`

### DIFF
--- a/git_hunk/_hunk.py
+++ b/git_hunk/_hunk.py
@@ -77,6 +77,10 @@ def _with_stable_ids(hunks: list[Hunk]) -> list[Hunk]:
     ]
 
 
+def split_file_diffs(diff_output: str) -> list[str]:
+    return re.split(r"(?=^diff --git )", diff_output, flags=re.MULTILINE)
+
+
 def extract_file_path(file_diff: str) -> str | None:
     first_line = file_diff.split("\n", 1)[0]
     # For non-renames git emits `diff --git a/<path> b/<path>` with both halves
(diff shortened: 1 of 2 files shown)


wkentaro:

**Verdict: recommend-merge**

Pure DRY refactor: extracts the duplicated `re.split(r"(?=^diff --git )", ..., flags=re.MULTILINE)` into a `split_file_diffs` helper, called from both `parse_diff` (`_hunk.py`) and `_extract_file_headers` (`_patch.py`).

- **Correctness**: behavior-preserving. The helper is byte-for-byte equivalent to both original call sites (same pattern, flags, default `maxsplit`). `re` remains in use in `_patch.py` (the `@@`-split), so no orphaned import.
- **Conventions**: the unprefixed name and the cross-module import from `_hunk` follow the existing precedent of `extract_file_path`, which is likewise unprefixed and imported into `_patch.py`.
- **CI**: green across all platforms and Python 3.10-3.13.

No rebase needed (branch is on current `main`, no file overlap). Mergeable, clean.

Issue #56, opened 2026-06-18
show --json with structured per-line hunk body


wkentaro:
> *This was generated by AI.*

## Problem

`git-hunk show <id>` renders the diff for terminal reading. Long content lines (e.g. long path/URL strings) hard-wrap across several visual lines, with the hunk line number printed only at the start of the logical line. For an agent this makes correlating a line number to its content error-prone — exactly the correlation that positional `-l` staging then depends on.

There is `list --json`, but no structured per-line representation of a hunk's body.

## Proposal

Add `show --json` (and/or `show <id> --json`) that emits each hunk with a structured, per-line body:

```json
{
  "id": "66279ce",
  "file": "path/to/file.py",
  "header": "@@ -94,37 +94,93 @@ ...",
  "lines": [
    {"n": 1, "op": " ", "content": "params=["},
    {"n": 2, "op": "+", "content": "    {\"resourceName\": \"...\"},"},
    {"n": 3, "op": "-", "content": "    \"old\","}
  ]
}
```

Where `n` is the 1-based position used by `-l`, and `op` is one of `" "` / `"+"` / `"-"`. This removes the terminal-wrapping ambiguity entirely and gives the agent a reliable map from line number to content.

## Notes

- Complements the content-based line selection request (a separate issue); even with pattern matching, a clean structured `show` is useful for the agent to decide what to match.
- One of a set of agent-usability ideas from the same session.

wkentaro:
> *This was generated by AI during triage.*

## Agent Brief

**Category:** feature
**Summary:** Add `show --json` emitting a structured per-line hunk body, per **ADR 0001 (schema v2)** — `docs/adr/0001-json-v2-hunk-schema.md` (PR #66).

**Current behavior:**
`show <id>` renders the diff for terminal reading; long lines hard-wrap, making line-number→content correlation (which `-l` staging depends on) error-prone. There is no structured per-line representation, and `list --json` currently carries a raw `diff` blob.

**Desired behavior (ADR §6, §7):**
- `show --json` emits each hunk as the **list fields plus** a `lines` array.
- A `lines[]` entry is `{n, op, content, no_newline?}`:
  - **`n`** — 1-based position within the hunk body. **Invariant:** must match `-l`'s existing numbering exactly — every body line counted, context included (`tests/e2e/line_selection_test.py` is the authority: `1=" a" 2="-b" 3="+B" ...`).
  - **`op`** — `" "` / `"+"` / `"-"`.
[...]

Issue #50, opened 2026-06-13
Function context shown twice in human list/applied display


wkentaro:
## What

For hunks that have a function/section context, the human `list` (and applied) display renders it twice:

- once inside `header`, because git emits `@@ … @@ def foo():`
- once via the separate `context_before` column (`git_hunk/_ui.py:53-54` and `171-172`)

e.g. `@@ -2,7 +2,7 @@ def foo():  def foo():`

## Why now

Surfaced while implementing #27 (PR #48), which deliberately left the display untouched since #27's acceptance criteria required no change to human-readable output. Now that `context_before` is also in `--json` (merged in bfadfaa), a JSON consumer that joins `header` + `context_before` sees the same redundancy, so it's worth resolving before #23 versions the schema.

## Options (maintainer to pick direction)

1. Render the bare `@@ -a,b +c,d @@` header (strip git's trailing context) and keep the dedicated `context_before` column as the single source.
2. Keep the full git header and drop the separate `context_before` column from the display.

Option 1 keeps the display column-aligned and treats `context_before` as the canonical field. No change to `--json` is needed either way; this is display-only.

## Source

- `git_hunk/_ui.py:53-54`, `git_hunk/_ui.py:171-172` (display)
- `git_hunk/_hunk.py:84-86` `_extract_context_before` (where the duplicate originates)

wkentaro:
> *This was generated by AI during triage.*

## Agent Brief

**Category:** bug
**Summary:** Function/section context is rendered twice (in `header` and again via `context_before`). Make `context_before` the single source of the section heading by normalizing the `header` field to the bare `@@ -a,b +c,d @@` range — fixing both the human display and the `--json` redundancy.

**Current behavior:**
git emits a hunk header like `@@ -11,7 +11,7 @@ def foo():`. The parser stores that whole string in the hunk's `header` field **and** separately extracts the trailing section heading (`def foo():`) into `context_before`. The human `list` and applied-line displays print both, producing visible duplication:

```
e314e6f  @@ -11,7 +11,7 @@ def foo():  def foo():  +1 -1
```

The same redundancy is now in `--json`: a consumer that reads both `header` and `context_before` sees the section heading twice.

**Desired behavior:**
[...]

PR #65, opened 2026-06-24
fix: byte-safe JSON for non-UTF-8 paths and content in list --json

### LINKED ISSUE #44: Emit byte-safe JSON for non-UTF-8 paths and content (lone surrogates break strict parsers)

wkentaro:
> *This was generated by AI during implementation.*

Closes #44.

`git-hunk` decodes git output with `surrogateescape`, so a non-UTF-8 byte in a
path or diff line becomes a lone surrogate in the Python `str`. `list --json`
serialized those with the default `ensure_ascii=True`, emitting a `\udcXX`
escape: `jq` silently rewrites it to U+FFFD (lossy) and strict parsers (Go
`encoding/json`, Rust `serde_json`) reject the whole document.

## Summary
- The git-derived hunk fields (`file`, `header`, `context_before`, `diff`) are
  now byte-safe objects following ripgrep's discriminated union: `{"text": "..."}`
  for valid UTF-8, `{"bytes": "<base64>"}` otherwise. Always an object, never a
  bare string, so consumers have one code path.
- `id`, `status`, `additions`, `deletions` stay plain (never git-derived).
- `schema_version` bumps to `2` (breaking shape change); README, the bundled
  `skills/core/SKILL.md`, and CHANGELOG document the new shape.

## Test plan
- [x] new tests: strict-parser acceptance + lossless base64 round-trip for a
  non-UTF-8 diff line, the `{"bytes"}` path for a non-UTF-8 filename (unit-level,
  since macOS rejects such names on disk), multi-byte UTF-8 staying `{"text"}`,
  and an envelope-contract assertion locking the v2 field shape
- [x] `uv run pytest` (195 passed)
- [x] manual smoke: `git-hunk list --json` on a repo with a `0xe9` byte in a
  diff line yields `{"bytes": ...}` that strict-parses and round-trips

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,11 @@ and this project adheres to
 - **Breaking:** `list --json` now wraps its output in a versioned envelope,
   `{"schema_version": 1, "hunks": [...]}`, instead of a bare array, so consumers
   can depend on a documented, versioned shape (#23).
+- **Breaking:** `list --json` emits the git-derived string fields (`file`,
+  `header`, `context_before`, `diff`) as byte-safe objects (`{"text": "..."}`
+  for UTF-8, `{"bytes": "<base64>"}` otherwise) instead of bare strings, and
+  bumps `schema_version` to 2. This keeps the output valid for strict JSON
+  parsers (Go, Rust) when a path or diff line carries non-UTF-8 bytes (#44).
 - README image paths are rewritten to absolute URLs so they render on PyPI.
 
 ### Fixed
--- a/README.md
+++ b/README.md
@@ -126,34 +126,42 @@ shape:
 
 ```json
 {
-  "schema_version": 1,
+  "schema_version": 2,
   "hunks": [
     {
       "id": "d161935",
-      "file": "src/main.py",
+      "file": { "text": "src/main.py" },
       "status": "unstaged",
-      "header": "@@ -10,3 +10,5 @@ def main():",
-      "context_before": "def main():",
+      "header": { "text": "@@ -10,3 +10,5 @@ def main():" },
+      "context_before": { "text": "def main():" },
       "additions": 2,
       "deletions": 0,
-      "diff": "..."
+      "diff": { "text": "..." }
     }
   ]
 }
 ```
 
-| Field            | Type   | Description                                                                                        |
-| ---------------- | ------ | -------------------------------------------------------------------------------------------------- |
-| `schema_version` | int    | Envelope version; bumped on any incompatible change to the shape below.                            |
-| `hunks`          | array  | The hunks (empty array when there are no changes).                                                 |
-| `id`             | string | Stable, content-based hunk id (7-char SHA-256 prefix); accepts prefixes.                           |
-| `file`           | string | Path of the changed file.                                                                          |
-| `status`         | string | One of `staged`, `unstaged`, `untracked`.                                                          |
-| `header`         | string | The hunk's `@@ ... @@` header, or a `Binary file (...)` / `Mode ...` label for whole-file changes. |
-| `context_before` | string | The function/section git names after the `@@` header; empty when git provides no context.          |
-| `additions`      | int    | Number of added lines.                                                                             |
-| `deletions`      | int    | Number of removed lines.                                                                           |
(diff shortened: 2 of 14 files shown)


wkentaro:

Finalized this own PR. No rebase needed: `main` has not advanced past the merge-base (0 commits, no file overlap), merge state is `CLEAN`/`MERGEABLE`.

- **Review gate** (`/code-review`, medium, report-only): no meaningful correctness or design findings. The core `_encode_arbitrary` probe (`s.encode("utf-8")` to detect the lone surrogates `surrogateescape` leaves, then base64 of the `surrogateescape` re-encode) is lossless and correct, including the mixed valid-UTF-8 + surrogate case. Only sub-threshold doc/test nits surfaced, matching existing repo conventions; `/review-fix` skipped.
- **Commits**: single, well-formed logical commit; no `/recommit` reshape needed.
- **Verify**: smoke of `list --json` confirms `schema_version: 2` and byte-safe `{"text": ...}` / `{"bytes": <base64>}` wrapping for `file`/`header`/`context_before`/`diff`.
- **CI**: green across all 13 jobs (lint + test on macOS/Ubuntu/Windows x py3.10-3.13); 195 tests pass locally.

No push this tick, so the existing green is authoritative.

wkentaro:

Superseding this in favor of #67. ADR 0001 (`docs/adr/0001-json-v2-hunk-schema.md`, merged in #66) converged #44 with #28/#40/#50/#56 into a single `schema_version: 2` contract, and this PR — written before that ADR — diverges from it on two central points that can't be reconciled within a standalone #44:

- It wraps **`header`** in the union, but the ADR makes `header` a **plain** bare `@@` range / `null` (the section heading and change-kind/mode labels move to `context_before` and typed fields, per #50/#40).
- It keeps and wraps the monolithic **`diff`** blob, but the ADR **removes** it in favor of a structured `lines[]` body in `show --json` (#56) — precisely because wrapping the blob means one non-UTF-8 byte base64-encodes the entire hunk body.

The genuinely good parts here — the `_encode_arbitrary` text/bytes helper (which matches the ADR exactly), the `schema_version` bump, and the non-UTF-8 tests — will be carried into the #67 implementation rather than discarded. Thanks; closing so the byte-safety lands as part of the one coordinated v2 change.

Issue #53, opened 2026-06-13
Support rename inventory and safe two-path mutations beyond the current rejection guard


wkentaro:
Audited on 2026-09-18 against [`5e46e0f`](https://github.com/wkentaro/git-hunk/tree/5e46e0f5dceb76a377dc06bda1421f7cd74ae688). This brief supersedes stale implementation assumptions in earlier comments; historical decisions remain linked below.

## Current behavior

The original silent-drop report is obsolete. `_get_hunks` now rejects detected rename/copy states before inventory output or mutation, including quoted paths; `tests/e2e/unsupported_file_state_test.py` covers staged and unstaged renames, quoted paths, mixed selections, and weak Git configuration. README and CONTEXT document this limitation and link here.

Rename support itself is still absent. Keep this open as a feature decision, not as an unfixed false-clean bug. A staged pure rename currently produces an explicit error directing the caller to Git.

## Decisions needed

- Define the rename's structured inventory representation and whether source/destination paths are separately exposed. Existing JSON uses `change_kind` A/D/M/T; R/C are reserved.
- Define safe stage, unstage, and discard semantics for both paths, including rename-with-content-change and interaction with unrelated edits.
- Decide scope relative to #22 (untracked additions) and #200 (copy detection); those are separate capabilities/policy choices.

## Acceptance after the contract is decided

- [ ] Pure renames become visible, addressable entries in plain and JSON output.
- [ ] Both paths are handled safely during mutation; unrelated changes survive, and unsupported combinations fail before mutation.
- [ ] Cover pure and edited renames, quoted/non-UTF-8 paths where supported, staged/unstaged state, dry-run, and mixed selections.
- [ ] Update the documented contract and add a towncrier fragment.

Do not remove the rejection guard before the complete supported contract exists.

wkentaro:
> *This was generated by AI during PR processing.*

Noticed while processing #68 (the quoted-path header fix for #30): the quoted-rename header form is a sub-case of this issue and is **more broken** than the description's "partially handled" note implies.

For an unquoted rename-with-content-change, the content hunk does list under the destination path (as described here). But when the path contains a tab/newline/backslash/double-quote, git emits a quoted header with **differing** halves:

```
diff --git "a/od\ttab.txt" "b/new\ttab.txt"
```

[...]

PR #68, opened 2026-06-24
fix: parse git's quoted, C-escaped diff headers

### LINKED ISSUE #30: Parse paths git always quotes (tab, newline, backslash, quote in filename)

wkentaro:
git always double-quotes and C-escapes a diff header whose path contains a tab, newline, backslash, or double-quote, regardless of `core.quotePath`. `extract_file_path` only matched the unquoted `diff --git a/… b/…` form, so such files were silently dropped from `list`/`stage`/`discard`. This adds a quoted-header branch that C-unescapes the logical path (`\t \n \\ \" \a \b \f \r \v` and octal `\NNN`); `git apply` already round-trips the raw quoted header, so patch emission is unchanged.

## Test plan
- [x] New parametrized e2e cases in `tests/e2e/quoted_paths_test.py` round-trip `list`/`stage`/`unstage`/`discard` for a tab, newline, backslash, and double-quote in the filename
- [x] Real CLI smoke: a tab-named file now appears in `git-hunk list` and stages to the correct content (was silently dropped before)
- [x] `make lint` and full `make test` (195 passed)

Closes #30

### DIFF
--- a/git_hunk/_hunk.py
+++ b/git_hunk/_hunk.py
@@ -81,8 +81,43 @@ def split_file_diffs(diff_output: str) -> list[str]:
     return re.split(r"(?=^diff --git )", diff_output, flags=re.MULTILINE)
 
 
+def _unquote_c_path(path: str) -> str:
+    C_ESCAPES: Final = {
+        "a": "\a",
+        "b": "\b",
+        "f": "\f",
+        "n": "\n",
+        "r": "\r",
+        "t": "\t",
+        "v": "\v",
+        "\\": "\\",
+        '"': '"',
+    }
+    chars = []
+    i = 0
+    while i < len(path):
+        if path[i] != "\\":
+            chars.append(path[i])
+            i += 1
+            continue
+        escape = path[i + 1]
+        if escape in C_ESCAPES:
+            chars.append(C_ESCAPES[escape])
+            i += 2
+        else:
+            chars.append(chr(int(path[i + 1 : i + 4], 8)))
+            i += 4
+    return "".join(chars)
+
+
 def extract_file_path(file_diff: str) -> str | None:
     first_line = file_diff.split("\n", 1)[0]
+    # git double-quotes and C-escapes the header when a path contains a tab,
+    # newline, backslash, or double-quote (regardless of core.quotePath). Both
+    # halves are identical for non-renames; decode the logical path.
+    m = re.match(r'diff --git "a/(.+)" "b/\1"$', first_line)
+    if m:
+        return _unquote_c_path(m.group(1))
     # For non-renames git emits `diff --git a/<path> b/<path>` with both halves
     # identical; the backreference resolves paths that contain " b/".
     m = re.match(r"diff --git a/(.+) b/\1$", first_line)
--- a/tests/e2e/quoted_paths_test.py
+++ b/tests/e2e/quoted_paths_test.py
@@ -1,3 +1,7 @@
+import sys
+
+import pytest
+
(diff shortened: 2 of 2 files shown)


wkentaro:

**Fix-attempt: Windows CI was red.**

The new `test_quoted_path_modified_file_round_trips` parametrizes filenames containing tab, newline, backslash, and `"` — all of which are illegal in Windows filenames. On `windows-latest` the test errored (`OSError: [Errno 22] Invalid argument` for tab/newline/quote, and `od\back.txt` resolved to `od/back.txt` for backslash). The production fix in `extract_file_path` / `_unquote_c_path` is correct; these characters simply cannot occur in a Windows path, so the test is genuinely Unix-only.

Applied a `@pytest.mark.skipif(sys.platform == "win32", ...)` guard matching the repo's existing convention (`binary_test.py`, `load_skills_test.py`), folded it into the single commit, and force-pushed. Local `ruff`/`ty`/`pytest` all green (195 passed on macOS); awaiting the re-triggered CI before a verdict.

follow-ups: the quoted-**rename** header form (`diff --git "a/old" "b/new"` with differing halves) still returns `None` and drops the file. That is out of scope for #30 (which targets the identical-halves non-rename case) and is captured under #53.

wkentaro:

**Verdict: recommend-merge.**

- CI is green across all 13 platform/version combos (lint + test on ubuntu/macos/windows × 3.10–3.13).
- Merge state is `CLEAN`; `main` advanced only in `.github/workflows/publish.yml` since the merge-base (disjoint from this PR's `git_hunk/_hunk.py` and `tests/e2e/quoted_paths_test.py`, 2 commits), so no rebase was needed — the existing green already reflects the merged result.
- A report-only `/code-review` pass found no in-scope correctness bugs: the `"a/(.+)" "b/\1"$` backreference can't over-capture, the octal branch advances correctly, and the `C_ESCAPES` table is complete for the chars git always quotes. No `/review-fix` needed.
- Smoke test confirmed: a file named `od\ttab.txt` round-trips through `list --json`.
- History is already a single clean, descriptive commit — no `/recommit` reshape needed.

follow-ups: #70 (`_unquote_c_path` mojibakes multibyte UTF-8 paths under default `core.quotePath` — out of scope for #30, decode-side, distinct from #44). Does not block this merge.

PR #71, opened 2026-06-24
feat: content/pattern-based line selection for partial hunk staging

### LINKED ISSUE #55: Content/pattern-based line selection for partial hunk staging
> *This was generated by AI.*

## Problem

Partial-hunk staging currently selects lines positionally: `stage <id> -l 115-119` / `-l ^115-119`. For an agent this is the one fragile spot in an otherwise content-addressed tool:

- It requires a `show` → read → translate-to-line-numbers round trip before every partial stage.
- Line numbers are positional, so they shift if the hunk changes between `show` and `stage`. Everything else in git-hunk is content-addressed (stable hashes); line selection is the exception.
- It does not match how an agent reasons about a diff. The agent is thinking "drop the new-resource entry", not "drop lines 115 through 119".

### Concrete example that motivated this

Migrating a `@pytest.mark.parametrize` block to `@evaluation.PytestMarkParametrize` produced a single full-rewrite hunk. To commit the migration without one newly-added entry, the only path was:

```bash
git-hunk show 66279ce            # read, find that the new entry is lines 115-119
git-hunk stage 66279ce -l ^115-119
```

The line-number lookup is pure overhead and is the part most likely to go wrong.

## Proposal

Add content/pattern-based line selection alongside positional `-l`:

```bash
git-hunk stage 66279ce --exclude-matching 'c1205_20260605'
git-hunk stage 66279ce --include-matching '"mark": "xfail"'
```

Semantics:

- Match against the content of added/removed lines within the named hunk (single-hunk only, like `-l`).
- `--include-matching` selects matching lines; `--exclude-matching` selects everything else. Composable with `stage`, `unstage`, `discard` (same as `-l`).
- Define behavior when a pattern matches zero lines (error, so a typo'd pattern never silently stages nothing/everything).

This lets the agent select by what the lines *are* rather than where they sit, removing the `show`/parse/translate round trip and the positional-staleness failure mode.

## Notes

- Positional `-l` should stay — it is the natural fit for a human reading the diff.
- This is the highest-leverage of a set of agent-usability ideas surfaced in the same session (machine-readable `show --json`, staged-file preview, per-commit validation hook); filing separately.

wkentaro:
Closes #55

## Summary
- Add `--include-matching` / `--exclude-matching` to `stage`, `unstage`, and `discard`, selecting changed (`+`/`-`) lines by content instead of by line number — so an agent can drop a line by what it contains without a `show` round trip or positional staleness.
- Patterns are literal substrings by default; `--regex` opts into Python regular expressions. Both flags are repeatable and OR'd, and case-sensitive.
- Zero matches is an error (nothing staged), so a typo'd pattern never silently stages nothing or everything.
- The matching flags are mutually exclusive with `-l` and with each other; `--regex` without a matching flag is rejected. Positional `-l` is unchanged.
- Pattern resolution reuses the existing line-filter machinery (`filter_hunk_lines`); the only new logic is mapping a pattern to the body line numbers `-l` already uses.

## Test plan
- [x] unit tests for `resolve_matching_lines` (literal/regex, OR, case-sensitivity, metacharacters, zero-match): `tests/unit/_lines/resolve_matching_test.py`
- [x] e2e tests across stage/unstage/discard incl. exclude, regex, mutual-exclusion, and `--regex`-without-pattern: `tests/e2e/pattern_selection_test.py`
- [x] regression test that empty `-l ""` still errors: `tests/e2e/error_test.py`
- [x] `uv run ruff format --check . && uv run ruff check . && uv run ty check` clean
- [x] full suite green (`uv run pytest`, 206 passed)
- [x] manual CLI smoke: exclude-matching drops the targeted line and leaves it unstaged; all error paths exit non-zero (1 content / 2 usage)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,11 @@ and this project adheres to
   change without touching the index or working tree (#25).
 - `context_before` field in `list --json`, exposing the function/section heading
   git names after the `@@` header (#27).
+- `--include-matching` / `--exclude-matching` for `stage`, `unstage`, and
+  `discard`, selecting changed lines by content instead of by line number
+  (literal substring by default, `--regex` for regular expressions; repeatable
+  and OR'd), so an agent can drop a line by what it contains without a
+  `show` round trip (#55).
 
 ### Changed
 
--- a/README.md
+++ b/README.md
@@ -99,11 +99,18 @@ git-hunk show --unstaged               # show all unstaged hunks
 git-hunk stage d161935                 # stage a hunk
 git-hunk stage d161935 a3f82c1         # stage multiple hunks
 git-hunk stage d161935 -l 3,5-7        # stage specific lines only
+git-hunk stage d161935 --exclude-matching debug    # stage all but lines containing "debug"
+git-hunk stage d161935 --include-matching xfail    # stage only lines containing "xfail"
 git-hunk unstage d161935               # move back to working tree
 git-hunk unstage d161935 -l 3,5-7      # unstage specific lines only
 git-hunk discard d161935               # restore from HEAD
 git-hunk discard d161935 -l ^3,^5-7    # discard excluding specific lines
 ```
+
+`--include-matching` / `--exclude-matching` select changed lines by content
+instead of line number (literal substring by default, `--regex` for regular
+expressions). Both are repeatable and OR'd, case-sensitive, and error if nothing
+matches. They are mutually exclusive with `-l` and with each other.
 
 ### Commit
 
--- a/git_hunk/_cli.py
+++ b/git_hunk/_cli.py
@@ -1,6 +1,7 @@
 import json
 import os
 import re
+from dataclasses import dataclass
 from dataclasses import replace
 from typing import Final
 
@@ -19,6 +20,7 @@ from ._hunk import Hunk
 from ._hunk import parse_diff
 from ._lines import filter_hunk_lines
 from ._lines import parse_line_spec
+from ._lines import resolve_matching_lines
 from ._patch import build_patch
 from ._skills import Skill
 from ._skills import load_skills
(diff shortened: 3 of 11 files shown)


wkentaro:

Rebased onto `main` to resolve a conflict in `git_hunk/_lines.py`: `main` had concurrently refactored the same file (introducing `_BodyLine` / `_select_body_lines` / `_render_body_lines` and a `strip_trailing_empty_lines` helper) while this branch added `resolve_matching_lines`. Resolution kept both:

- New `_body_lines(hunk)` now delegates to `strip_trailing_empty_lines` instead of duplicating its trailing-empty-strip loop.
- `filter_hunk_lines` and `resolve_matching_lines` share `_body_lines`, so changed-line numbering stays consistent between selection and filtering.
- Folded `main`'s three new no-newline tests in `tests/unit/_lines/filter_test.py` onto the branch's `make_hunk` conftest fixture (the old module-level `_make_hunk` helper was removed in this branch).

After resolution: `ruff format --check`, `ruff check`, `ty check` clean; `pytest` 219 passed. A report-only code review of the merge resolution found no meaningful issues. CLI smoke confirmed `stage --exclude-matching` stages the non-matching changed line and leaves the matching one unstaged.

Force-pushed; CI is running (no prior CI run existed on this branch). Verdict will follow once checks complete.

wkentaro:

CI completed green on all 13 checks (lint + test across ubuntu/macos/windows × Python 3.10-3.13) for the rebased head `887911a`. No new commits since the prior verdict comment, so the finalize state holds:

- Rebase onto `main` already done last tick (conflict in `_lines.py` resolved, both sides kept).
- Report-only code review found no meaningful issues; `/review-fix` skipped.
- History is a single clean commit.
- CLI smoke confirmed `stage --exclude-matching` behavior.

Verdict: **recommend-merge**.

PR #86, opened 2026-07-06
test: cover no-newline marker numbering in resolve_matching_lines

wkentaro:
`resolve_matching_lines` skips `\ No newline at end of file` markers without
counting them toward its 1-based line numbering, keeping it aligned with `-l`'s
numbering (context counted, markers excluded). The sibling `filter_hunk_lines`
path has three dedicated marker tests; this path had none. These two lock in the
numbering for a marker interleaved mid-hunk and one trailing at the hunk end.

## Test plan
- [x] `uv run pytest tests/unit/_lines/resolve_matching_test.py` (11 pass)
- [x] `uv run pytest` full suite (249 pass), `uv run ruff check .`, `uv run ty check`

### DIFF
--- a/tests/unit/_lines/resolve_matching_test.py
+++ b/tests/unit/_lines/resolve_matching_test.py
@@ -2,6 +2,7 @@ from collections.abc import Callable
 
 import pytest
 
+from git_hunk._hunk import NO_NEWLINE_MARKER
 from git_hunk._hunk import Hunk
 from git_hunk._lines import resolve_matching_lines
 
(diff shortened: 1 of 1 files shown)


wkentaro:

Test-only PR adding coverage for the no-newline-marker branch in `resolve_matching_lines` (the `is_no_newline_marker(line) -> continue` path that skips the marker without incrementing the line counter). This branch had no direct test before.

- Reviewed both assertions against the source: the interleaved case keeps `+new` at line 3 (marker between line 2 and 3 doesn't consume a number), and the trailing case keeps `+D` at line 6. Both correct.
- Uses the existing `make_hunk` fixture and imports `NO_NEWLINE_MARKER` from source rather than hardcoding the literal.
- Branch is current with `main` (no rebase needed), `/code-review` at high effort found nothing, all 7 CI checks green.

Test-only, so no CHANGELOG entry is required (the convention is for user-facing changes). Ready for maintainer merge.

PR #93, opened 2026-07-08
refactor: dedup staged/unstaged hunk selection in list/show

wkentaro:
`cmd_list` and `cmd_show` each carried a byte-for-byte copy of the same
`--staged`/`--unstaged` mutual-exclusion check and three-way (staged / unstaged /
both) hunk selection. This extracts that into one `_collect_hunks(*, staged, unstaged,
files, include_untracked)` helper. The two callers' only real differences stay explicit
at the call site: `list` passes the file filter and includes untracked entries in the
default case; `show` passes neither.

Behavior-preserving; no `--json` schema or CLI surface change, so no CHANGELOG entry.

## Test plan
- [x] `make test` — 256 passed
- [x] `make lint` — ruff format/check, ty, taplo, mdformat, yamlfix, typos all clean
- [x] Drove the real CLI in a temp repo: `list` (default shows staged+unstaged+untracked),
      `list --staged` / `--unstaged` (filtered, untracked omitted), `show` (default,
      no untracked), `show --staged`, and `list/show --staged --unstaged` (both error
      `cannot use --staged and --unstaged together`, exit 1)

### DIFF
--- a/git_hunk/_cli.py
+++ b/git_hunk/_cli.py
@@ -339,6 +339,31 @@ def _get_untracked_entries(files: list[str] | None = None) -> list[Hunk]:
     ]
 
 
+def _collect_hunks(
+    *,
+    staged: bool,
+    unstaged: bool,
+    files: list[str] | None,
+    include_untracked: bool,
+) -> list[Hunk]:
+    if staged and unstaged:
+        raise CliError("cannot use --staged and --unstaged together")
+
+    if staged:
+        hunks, _ = _get_hunks(staged=True, files=files)
+        return hunks
+    if unstaged:
+        hunks, _ = _get_hunks(staged=False, files=files)
+        return hunks
+
+    hunks_staged, _ = _get_hunks(staged=True, files=files)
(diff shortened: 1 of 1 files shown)


wkentaro:

Reviewed the diff (`git_hunk/_cli.py`): a clean, behavior-preserving extraction of the shared staged/unstaged hunk-selection logic into `_collect_hunks`. Confirmed each call site preserves prior behavior:

- `cmd_list` keeps untracked entries in the default case (`include_untracked=True`).
- `cmd_show` keeps its original behavior of excluding untracked (`include_untracked=False`).
- The `--staged`/`--unstaged` mutual-exclusion guard and `files=None` semantics are preserved.

Report-only `/code-review` surfaced no meaningful findings. History is already a single logical commit, so no reshaping was needed, and as a behavior-preserving refactor there's nothing to smoke-test. No rebase was needed (main gained 0 commits since the merge-base; merge state `CLEAN`), so the existing green CI is authoritative: all 7 checks pass.

Merging remains your call.

PR #95, opened 2026-07-08
fix: normalize CLI path arguments before matching files

wkentaro:
`git-hunk list ./foo.py` silently dropped an untracked `foo.py`. The untracked
filter compared `git ls-files`' canonical paths against the raw CLI argument, so
a `./`-prefixed (or otherwise non-canonical) path never matched, even though the
tracked-file half of the same command resolves it fine through git's pathspec.

The fix routes every CLI path argument through one `_normalize_path_arg` helper,
shared by `_select_hunks` and the untracked filter. It collapses `./` and
translates the OS-native separator via `posixpath` (git reports forward-slash
paths on every platform, so `os.path.normpath` would rewrite `/` to `\` and
mismatch on Windows).

## Test plan
- [x] `tests/e2e/list_test.py::test_list_file_filter_untracked_normalizes_path` reproduces the bug (fails without the fix)
- [x] `tests/e2e/path_shorthand_test.py::test_subdirectory_path` covers subdir path matching
- [x] `tests/unit/_cli/normalize_path_arg_test.py` covers the helper, incl. a monkeypatched Windows-separator case
- [x] `make test` (263 passed) and `make lint`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -55,5 +55,9 @@ and this project adheres to
 - Reject an empty `--include-matching` / `--exclude-matching` pattern, which
   previously matched every line and silently selected the whole hunk, so an
   accidentally-empty pattern now errors like an empty `-l` spec (#87).
+- Normalize file-path arguments before matching, so `git-hunk list ./foo.py`
+  surfaces an untracked `foo.py` instead of silently dropping it, and path
+  arguments resolve consistently across `list`/`stage`/`unstage`/`discard`
+  regardless of the platform's path separator (#95).
 
 [unreleased]: https://github.com/wkentaro/git-hunk/compare/v0.2.0...HEAD
--- a/git_hunk/_cli.py
+++ b/git_hunk/_cli.py
@@ -1,5 +1,6 @@
 import json
 import os
+import posixpath
 import re
 import stat
 from dataclasses import dataclass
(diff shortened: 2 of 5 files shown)


wkentaro:

Finalized this tick:

- **Rebase:** not needed. `main` gained 0 commits since the merge-base and merge state is CLEAN/MERGEABLE, so the existing green CI already reflects the merged result.
- **Code review** (medium, report-only): clean. `_normalize_path_arg` is the only Python-side path-string comparison against git's forward-slash paths; other path args are handed to git as pathspecs, which normalize `./` and separators themselves, so no other call site needs the helper. The Windows fix is real: the old `os.path.normpath` rewrote `/` to `\` and would never equal git's paths, and `_get_untracked_entries` previously did no normalization (raw `p in files`), silently dropping `./foo.py`.
- **Commits:** single logical commit; no reshaping needed.
- **Verify:** smoke-tested the fix end-to-end — `git-hunk list --json ./untracked.py` now surfaces the untracked file (`status: untracked`), matching the no-prefix control. Full suite: 263 passed.
- **CI:** all 7 checks green (lint + test across ubuntu/macos/windows on 3.10 and 3.14). No push this tick, so the existing green is authoritative.

Sound, covered, and green. Recommending merge for maintainer review.

PR #102, opened 2026-07-10
refactor: dedup the @@-hunk-header split regex into a shared helper

wkentaro:
The lookahead split `re.split(r"(?=^@@)", ...)` that separates a file diff into its
pre-hunk header and per-`@@` sections was hand-rolled in two modules: `parse_diff`
(`_hunk.py`) and `_extract_file_headers` (`_patch.py`). This extracts it as
`split_at_hunk_headers`, a sibling to the existing `split_file_diffs`, so the
hunk-boundary rule lives in one place.

Pure refactor, no behavior change. The helper takes an optional keyword-only `maxsplit`
so `_patch.py` keeps its original bounded (`maxsplit=1`) split, which only needs the
header before the first `@@`.

## Test plan
- [x] `uv run pytest -q` (264 passed), `ruff check`, `ruff format --check`, `ty check`
- [x] Smoke: `git-hunk list` (two-hunk file) then `stage <first-id>` stages only the
  first hunk, exercising both the full split (`parse_diff`) and the bounded split
  (`build_patch`)

### DIFF
--- a/git_hunk/_hunk.py
+++ b/git_hunk/_hunk.py
@@ -126,6 +126,10 @@ def split_file_diffs(diff_output: str) -> list[str]:
     return re.split(r"(?=^diff --git )", diff_output, flags=re.MULTILINE)
 
 
+def split_at_hunk_headers(file_diff: str, *, maxsplit: int = 0) -> list[str]:
+    return re.split(r"(?=^@@)", file_diff, maxsplit=maxsplit, flags=re.MULTILINE)
+
+
 def _unquote_c_path(path: str) -> str:
     C_ESCAPES: Final = {
         "a": "\a",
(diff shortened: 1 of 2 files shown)


wkentaro:

Behavior-preserving refactor that extracts the duplicated `re.split(r"(?=^@@)", ...)` hunk-header split into a shared `split_at_hunk_headers` helper, called from `parse_diff` (maxsplit=0) and `_extract_file_headers` (maxsplit=1).

- Both call sites preserve the original semantics exactly; no remaining `(?=^@@)` regex sites remain outside the helper.
- Dropping `import re` from `_patch.py` leaves no orphan (`re.` is unused there now).
- `split_at_hunk_headers` is public, matching the sibling `split_file_diffs` convention.
- CI green across all platforms; lint + 264 tests pass locally. Not user-facing, so no CHANGELOG entry needed.

PR #103, opened 2026-07-10
fix: report untracked files with repo-root-relative paths

wkentaro:
## Summary
- `git ls-files --others` returned paths relative to the current working directory, while `git diff` always reports repo-root-relative paths. Run from a subdirectory, `git-hunk list` emitted untracked hunks whose `file` was on a different path basis than the staged/unstaged hunks in the same envelope.
- Add `--full-name` so untracked paths are repo-root-relative too, and let git resolve the file-path filter as a pathspec (`get_untracked_files(files=...)`), mirroring `get_diff(files=...)` instead of matching paths by hand on a mismatched basis.
- The working-tree mode is looked up via the repo root so `lstat` still finds the file from a subdirectory.

## Test plan
- [x] Added `tests/e2e/subdirectory_test.py`: untracked `file` matches the tracked basis from a subdirectory, and filtering an untracked file by a relative arg from a subdirectory still matches.
- [x] Smoke: real `git-hunk list --json` run from `pkg/sub/` reports `unstaged pkg/app.py` and `untracked pkg/sub/fresh.py` (both repo-root-relative).
- [x] `uv run pytest -q` (266 passed), `uv run ruff check .`, `uv run ty check`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -104,5 +104,8 @@ and this project adheres to
   hunk as carrying a usable `id`: an `untracked` entry's `id` is empty, so no
   git-hunk command can address it, and plain `list` renders it as a bare path
   (#162).
+- Report untracked files with repo-root-relative paths, matching tracked hunks,
+  so running `git-hunk` from a subdirectory no longer emits an untracked `file`
+  on a different path basis than the staged/unstaged hunks beside it (#103).
 
 [unreleased]: https://github.com/wkentaro/git-hunk/compare/v0.2.0...HEAD
--- a/git_hunk/_cli.py
+++ b/git_hunk/_cli.py
@@ -15,6 +15,7 @@ from ._git import apply_patch
 from ._git import commit
 from ._git import discard_files
 from ._git import get_diff
+from ._git import get_repo_root
 from ._git import get_untracked_files
 from ._git import is_git_repo
 from ._git import stage_files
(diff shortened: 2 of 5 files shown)


wkentaro:

Finalized this tick — recommending merge.

- **Rebase:** skipped. `main` gained 0 commits since the merge-base and merge state is CLEAN, so the existing green CI already reflects the merged result.
- **Review:** report-only `/code-review` (medium). No meaningful findings. The reviewers surfaced that untracked-file filtering now uses git pathspec semantics (glob metacharacters, trailing-slash, directory expansion) instead of the previous exact-string match. That is the intended consistency: `get_diff(files=...)` already passes `files` as a raw pathspec, and `list` now hands the same raw args to both the tracked and untracked paths, so untracked filtering is byte-for-byte identical to tracked. Not a regression to fix. (Minor, pre-existing and intentional: a literal filename containing `[` `]` `*` `?` is interpreted as a glob for both tracked and untracked args.)
- **Commits:** already a single clean logical commit; `/recommit` was a no-op.
- **Verify:** ran `git-hunk list --json` from a subdirectory against a scratch repo. Tracked (`sub/tracked.py`) and untracked (`sub/untracked.py`) hunks report on the same repo-root-relative basis, and the untracked `b_mode` (`100644`) resolves correctly via the `posixpath.join(root, p)` fix.
- **CI:** all 7 checks green (lint + test across ubuntu/macos/windows on 3.10 and 3.14). No push happened this tick, so the existing green is authoritative.

PR #105, opened 2026-07-10
fix: stop commit --help advertising options it rejects

wkentaro:
`git-hunk commit --help` advertised `--include-matching`, `--exclude-matching`,
and `--regex`, but `commit` only ever declared `-m`, `-l`, and `-h`. Passing any
of those flags failed with a raw Click `Error: No such option`, bypassing the
tool's own styled errors. The cause was `HELP_COMMIT` reusing the shared
`_LINE_OPT_ROW` (which carries the matching options) instead of just the `-l`
row. This extracts the `-l` row into `_LINE_SELECT_ROW` and has `HELP_COMMIT`
reference only that, so the help lists exactly the options `commit` accepts. The
`stage`/`unstage`/`discard` help is unchanged (still uses the full row).

## Test plan
- [x] `tests/e2e/error_test.py`: `commit --help` omits the three matching
  options and still shows `-m`/`-l`; `commit --include-matching` errors; and a
  guard that `stage --help` still advertises the matching options.
- [x] `uv run pytest -q` (267 passed), `uv run ruff check .`, `uv run ty check`
- [x] Drove `commit --help`, `commit --include-matching`, and `commit -l 1 -m`
  end-to-end.

Possible follow-up (not in this PR): the `HELP_*` strings are hand-kept in sync
with each command's `@click.option` declarations, so this drift class could
recur for another command. A parametrized test walking each command's real
Click params against its help string would prevent it, but parsing options out
of Rich-markup text is fiddly enough to be its own change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,9 @@ and this project adheres to
 
 ### Fixed
 
+- `commit --help` no longer advertises `--include-matching`,
+  `--exclude-matching`, and `--regex`, which the command never accepted; the
+  help now lists only the options `commit` actually supports (#105).
 - Preserve `\ No newline at end of file` markers so staging the last line of a
   file without a trailing newline no longer fails or silently stages nothing
   (#9).
--- a/git_hunk/_ui.py
+++ b/git_hunk/_ui.py
@@ -217,9 +217,12 @@ def print_version(version: str) -> None:
     _err().print(f"git-hunk [dim]{version}[/dim]")
 
 
-_LINE_OPT_ROW = """\
+_LINE_SELECT_ROW = """\
   [bold cyan]-l[/bold cyan] [cyan]<lines>[/cyan]  Select specific lines within a hunk (requires exactly one hunk)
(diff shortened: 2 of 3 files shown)


wkentaro:

Reviewed and finalized this tick.

- **Fix is correct.** `cmd_commit` in `git_hunk/_cli.py` declares only `-m`, `-l`, `-h/--help` (and hardcodes `include_matching=()`, `exclude_matching=()`, `regex=False`), so it never accepted `--include-matching`/`--exclude-matching`/`--regex`. The old `commit --help` advertised them anyway; splitting `_LINE_OPT_ROW` into `_LINE_SELECT_ROW` (the `-l` row) fixes that.
- **No regression on other help.** `stage`/`unstage`/`discard` help still list all matching options (verified by rendering the real help output).
- **Verified end-to-end.** `commit --help` now shows only `-m`/`-l`; `stage --help` still shows the matching options. `tests/e2e/error_test.py` green (16 passed).
- **CI green** across lint + all 6 test matrix jobs; merge state CLEAN, no rebase needed (main has not advanced past the merge-base).
- **Code review:** clean, no meaningful findings.

Ready for maintainer review and merge.

PR #106, opened 2026-07-10
refactor: dedup the shared line-selection options for stage/unstage/discard

wkentaro:
`stage`, `unstage`, and `discard` each repeated the same 7-line click
option/argument stack (`-l`, `--include-matching`, `--exclude-matching`,
`--regex`, `--dry-run`, `-h/--help`, `targets`). Extract it into one
`_add_patch_selection_options` decorator so a future option added or renamed
in one command can't silently drift from the other two, the same failure class
fixed on the help-string side in #105.

Behavior-preserving: click binds parameters by dest name, so the resulting
`click.Command.params` order is unchanged and `--help` output is byte-identical
for all three commands. `cmd_commit` keeps its own smaller option set (it isn't
a subset of this stack) and is left untouched.

## Test plan
- [x] `uv run pytest -q` — 264 passed
- [x] `uv run ruff check` / `uv run ty check` — clean
- [x] `git-hunk {stage,unstage,discard} -h` output diffed against `main` — byte-identical

### DIFF
--- a/git_hunk/_cli.py
+++ b/git_hunk/_cli.py
@@ -3,6 +3,7 @@ import os
 import posixpath
 import re
 import stat
+from collections.abc import Callable
 from dataclasses import dataclass
 from dataclasses import replace
 from typing import Final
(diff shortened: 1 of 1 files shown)


wkentaro:

Pure internal refactor: the three identical `@click.option`/`@click.argument` decorator stacks on `stage`/`unstage`/`discard` collapse into one `_add_patch_selection_options` helper. `reversed(options)` reproduces the original bottom-up decorator stacking order exactly (argument first, then `-h` → `-l`), so the wiring is behavior-preserving.

- `/code-review` (medium): no findings.
- All 264 tests pass; ruff clean.
- Smoke: `stage`/`unstage`/`discard -h` each still advertise `-l`, `--include-matching`, `--exclude-matching`, `--regex`, `--dry-run`.
- CI: all 7 checks green.
- No CHANGELOG entry needed (internal refactor, no user-facing change).

Branch untouched this tick (no rebase needed — 0 commits behind main; no fixes required).

PR #109, opened 2026-07-11
docs: correct discard as restoring from the index, not HEAD

wkentaro:
`git-hunk discard` reverse-applies the working-tree-vs-index diff (and runs `git restore -- <file>` for whole-file changes), so it restores a hunk to the staged content, not to HEAD. The `--help` text, README, and bundled skill doc all said "restore from HEAD", which only holds when the file has no staged sibling; with a staged edit in the same file, discarding an unstaged hunk leaves the staged edit intact rather than collapsing the file to the last commit.

Corrects all six spots and adds a regression test for the index-vs-HEAD distinction, which was previously uncovered.

## Test plan
- [x] `uv run pytest -q` — new `test_discard_restores_from_index_not_head` plus full suite (266 passing)
- [x] `uv run ruff check . && uv run ty check`
- [x] `git-hunk discard --help` renders "restore from the index"

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -83,5 +83,9 @@ and this project adheres to
   and `show`, matching every other conflicting-flags error (#117).
 - Match hunk ids case-insensitively, so an uppercased id like `git-hunk show 713B7B9` resolves like git's own object-id lookup instead of failing with a
   misleading "not found" / "no changed file matches" error (#150).
+- Correct the `discard` help, README, and skill docs, which described it as
+  restoring "from HEAD" when it restores from the index: discarding an unstaged
+  hunk reverts it to the staged content, leaving a staged sibling edit intact
+  (#109).
 
 [unreleased]: https://github.com/wkentaro/git-hunk/compare/v0.2.0...HEAD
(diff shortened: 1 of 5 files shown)


wkentaro:

Reviewed the change end-to-end:

- **Docs accuracy confirmed against the code.** `cmd_discard` runs the unstaged (index-vs-worktree) diff and applies it with `git apply --reverse` (no `--cached`), and the whole-file fallback uses `git restore -- <file>` — both restore the working tree to the **index** content, not HEAD. The old "restore from HEAD" wording only held when a file had no staged sibling; the corrected wording is accurate. All six doc spots (README, three in `_ui.py`, two in `SKILL.md`) are fixed, and no stale "from HEAD" discard reference remains.
- **Regression test is a real discriminator.** `test_discard_restores_from_index_not_head` stages `STAGED/line2/line3`, leaves `STAGED/line2/UNSTAGED` unstaged, discards the unstaged hunk, and asserts the result is the staged content — which fails if discard restored from HEAD. Genuinely proves index-not-HEAD, not just any-restore.
- **CI green** across all 7 checks (lint + test matrix). No rebase needed: branch is 0 commits behind `main`, merge state clean.

Commit history is already clean: docs correction, then its regression test. No follow-ups.

PR #110, opened 2026-07-11
test: cover parse_diff skipping a headerless file-diff block

wkentaro:
`parse_diff` skips a split file-diff block whose first line is not a `diff --git`
header (so `extract_file_path` returns `None`), before it can emit a hunk with no
path (`git_hunk/_hunk.py:251-253`). That skip branch was untested. This adds a
direct test: a leading block carrying an `@@` body (which makes the skip
load-bearing, i.e. without the guard a spurious pathless hunk would be counted)
followed by a well-formed file diff, asserting only the real hunk survives.

## Test plan
- [x] `uv run pytest tests/unit/_hunk/parse_diff_test.py` (9 passed)
- [x] confirmed the test fails if the `filepath is None` skip is removed (mutation: 2 hunks instead of 1)
- [x] `make lint`

### DIFF
--- a/tests/unit/_hunk/parse_diff_test.py
+++ b/tests/unit/_hunk/parse_diff_test.py
@@ -239,3 +239,27 @@ def test_delete_then_unrelated_add_is_not_merged_as_typechange() -> None:
     assert hunks[1].file == "b.txt"
     assert hunks[1].change_kind == "A"
     assert hunks[1].additions == 1
+
+
+def test_chunk_without_diff_header_is_skipped() -> None:
+    # A chunk with no `diff --git` header has no extractable path, so parse_diff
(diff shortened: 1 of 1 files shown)


wkentaro:

Finalized: **recommend-merge**.

- Single test-only commit; adds coverage for `parse_diff` skipping a leading chunk that has no `diff --git` header (the `filepath is None` branch at `git_hunk/_hunk.py:251`).
- Up to date with `main` (0 commits behind base, merge state CLEAN) — no rebase needed.
- `/code-review` (medium) surfaced no meaningful findings; test follows repo conventions.
- CI green across all lint/test matrix jobs.

No follow-up or idea issues warranted.

PR #112, opened 2026-07-11
refactor: reuse whole_file_hunk factory for untracked entries

wkentaro:
`_get_untracked_entries` in `_cli.py` hand-wrote the same 11-field whole-file
`Hunk` literal (`id=""`, `header=None`, `context_before=None`, `additions=0`,
`deletions=0`, `diff=""`, …) that `_hunk.py`'s `_whole_file_hunk` already
factors, so adding a field to `Hunk` would silently drift the untracked path.
Promote the factory to public `whole_file_hunk` with a `status` parameter (the
untracked case passes `status="untracked"`; the three in-module callers keep the
`"unstaged"` default) and reuse it. Output is byte-identical.

## Test plan
- [x] `make test` (264 passed) — existing untracked-entry e2e coverage unchanged
- [x] `make lint` (ruff format + check, ty, taplo, mdformat, yamlfix, typos)
- [x] Smoke: `git-hunk list --json` on an untracked file still yields
      `status=untracked, change_kind=A, a_mode=null, b_mode=100644, binary=false`

### DIFF
--- a/git_hunk/_cli.py
+++ b/git_hunk/_cli.py
@@ -21,6 +21,7 @@ from ._git import unstage_files
 from ._hunk import Hunk
 from ._hunk import is_whole_file_hunk
 from ._hunk import parse_diff
+from ._hunk import whole_file_hunk
 from ._lines import filter_hunk_lines
 from ._lines import parse_line_spec
 from ._lines import resolve_matching_lines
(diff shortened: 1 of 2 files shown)


wkentaro:

Behavior-preserving refactor. `_whole_file_hunk` is renamed to public `whole_file_hunk` and gains a `status: str = "unstaged"` parameter so `_get_untracked_entries` in `_cli.py` can reuse the factory instead of constructing a `Hunk` inline.

Finalize checks:
- **Behavior preserved.** `Hunk.status` already defaults to `"unstaged"`, so the three existing `parse_diff` call sites (which never pass `status`) are unchanged; the untracked call still yields a field-identical `Hunk` (`status="untracked"`).
- **Naming consistent.** Dropping the leading underscore matches the module convention: every non-underscore top-level function in `_hunk.py` (`parse_diff`, `extract_file_path`, etc.) is imported cross-module; `whole_file_hunk` now is too.
- **Green:** lint, `ty`, and 264 tests pass locally; all CI checks green. No rebase needed (branch is based on current `main`, 0 commits diverged), so the existing green reflects the merged result.

No follow-ups or spin-off issues.

PR #113, opened 2026-07-12
refactor: dedup the context+stats tail in hunk-line rendering

wkentaro:
`_print_hunk_line` and `print_applied` each carried a byte-identical four-line
tail that appends the `context_before` heading, a two-space separator, and the
`+/-` stats to the rendered hunk line. Extract it into
`_append_context_and_stats` so the two renderers can no longer drift apart when
that trailing format changes. Pure refactor; rendered output is unchanged.

## Test plan
- [x] `uv run ruff check .` / `uv run ruff format --check` / `uv run ty check` clean
- [x] Rendered `git-hunk list` and `git-hunk stage --dry-run` on a hunk with a
  `def foo():` heading; both show the heading + stats identically to `main`

### DIFF
--- a/git_hunk/_ui.py
+++ b/git_hunk/_ui.py
@@ -55,7 +55,10 @@ def _header_text(hunk: Hunk) -> str:
     return hunk.header if hunk.header is not None else _whole_file_label(hunk)
 
 
-def _append_stats(text: Text, hunk: Hunk) -> None:
+def _append_context_and_stats(text: Text, hunk: Hunk) -> None:
+    if hunk.context_before:
+        text.append(f"  {_safe(hunk.context_before)}", style="dim italic")
+    text.append("  ")
     if hunk.additions:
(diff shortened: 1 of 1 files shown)


wkentaro:

Finalized this tick. Verdict: **recommend-merge**.

- Branch is current with `main` (0 commits behind the merge-base), so the existing green CI reflects the merged result; no rebase.
- Cheap `/code-review` pass over the diff: no findings. The change extracts `_append_context_and_stats` from two byte-identical call sites in `_print_hunk_line` and `print_applied`; the helper reproduces the original sequence exactly, so it is behavior-preserving.
- Single clean commit already; `/recommit` unnecessary. Pure rendering refactor with no behavior change, so `/verify` skipped.
- All CI green (lint + test matrix across ubuntu/macos/windows on 3.10 and 3.14).

PR #116, opened 2026-07-12
refactor: drop dead unbared-header fallback in _bare_header

wkentaro:
`_bare_header`'s `else at_line` fallback was dead: `parse_diff` splits each file diff on `(?=^@@)` and feeds every segment's first line, which — given git's pinned `-U3` output — always starts with a well-formed `@@ -a,b +c,d @@` header. The regex therefore always matches, and the fallback would only ever fire to silently return the *unbared* line (git's section heading included) as the JSON `header` field.

Replace it with a loud `ValueError` on non-match, mirroring `filter_hunk_lines` in `_lines.py`, which handles the same header shape and fails loudly rather than mislabeling. The raise is an unreachable invariant guard (documented as such), not a new runtime error path; the CLI parse path has no reachable `ValueError`, so no boundary handling is added.

## Test plan
- [x] tests added: `tests/unit/_hunk/bare_header_test.py` (heading strip, already-bare, single-line ranges, and the non-match raise)
- [x] `uv run pytest -q` (268 passed)
- [x] `uv run ruff check .` / `uv run ty check`

### DIFF
--- a/git_hunk/_hunk.py
+++ b/git_hunk/_hunk.py
@@ -187,7 +187,12 @@ def _bare_header(at_line: str) -> str:
     # "@@ -1,3 +1,3 @@ def foo():" -> "@@ -1,3 +1,3 @@" (strip git's heading; the
     # heading is carried separately in context_before).
     m = re.match(r"(@@ -\d+(?:,\d+)? \+\d+(?:,\d+)? @@)", at_line)
-    return m.group(1) if m else at_line
+    if not m:
+        # Unreachable for real input: parse_diff only ever passes a line that the
+        # (?=^@@) split guarantees starts with git's well-formed @@ header. Raise
+        # rather than silently mislabel if that invariant is ever broken.
+        raise ValueError(f"cannot parse hunk header: {at_line}")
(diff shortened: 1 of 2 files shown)


wkentaro:

**Verdict: recommend-merge**

Reviewed the diff at medium effort; no meaningful findings.

- The change replaces the silent `return at_line` fallback in `_bare_header` with an explicit `ValueError`. I traced the caller: `header_line` comes from `re.split(r"(?=^@@)", ...)` (line 288), and in standard unified-diff output every hunk-body line carries a `+`/`-`/space prefix, so only genuine `@@ -N +N @@` headers reach `_bare_header`. The raised branch is unreachable for real git output (combined `@@@` diffs do not occur for working-tree/index diffs), so this converts a dead fallback into a fail-loud invariant check without changing observable behavior.
- Tests cover all four cases (strip heading, bare-unchanged, single-line ranges, raise-on-mismatch).
- CI: lint + all 6 test matrix jobs green. Branch is CLEAN and 0 commits behind main, so no rebase was needed.

PR #117, opened 2026-07-12
fix: show a Usage hint on the --staged/--unstaged conflict error

wkentaro:
The `--staged`/`--unstaged` mutual-exclusion error is raised in the shared
`_collect_hunks` helper (used by both `list` and `show`), and was the only
conflicting-flags error in the CLI that omitted the `usage=` hint every sibling
error already carries (`stage`/`unstage`/`discard`/`commit`/`skills`). The user
got no `Usage:` line and, because `usage` was absent, the error exited 1 instead
of the 2 those sibling errors use.

This threads each command's usage string through `_collect_hunks` (`list` →
`USAGE_LIST`, `show` → `USAGE_SHOW`) and extracts `USAGE_LIST` so `HELP_LIST`
references the constant like every other `HELP_*` block instead of inlining a
duplicate literal. The `usage` parameter is required (no default) so a future
third caller can't silently regress.

## Test plan
- [x] Extended `test_list_staged_and_unstaged_mutual_exclusion` and
  `test_show_staged_and_unstaged_together_errors` to assert the per-command
  `Usage:` line
- [x] `make lint`
- [x] Drove both commands: `git-hunk list --staged --unstaged` and
  `git-hunk show --staged --unstaged` now print their command-specific usage and
  exit 2; `list --help` renders byte-identically from the extracted constant

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -76,5 +76,7 @@ and this project adheres to
 - Report a git failure on a bad path argument (e.g. `git-hunk list ':(bogus)x'`)
   as a clean `error:` message instead of crashing with a raw Python traceback
   (#107).
+- Show a `Usage:` hint on the `--staged`/`--unstaged` conflict error for `list`
+  and `show`, matching every other conflicting-flags error (#117).
 
 [unreleased]: https://github.com/wkentaro/git-hunk/compare/v0.2.0...HEAD
--- a/git_hunk/_cli.py
+++ b/git_hunk/_cli.py
@@ -38,6 +38,8 @@ from ._ui import HELP_UNSTAGE
 from ._ui import USAGE
 from ._ui import USAGE_COMMIT
 from ._ui import USAGE_DISCARD
+from ._ui import USAGE_LIST
+from ._ui import USAGE_SHOW
 from ._ui import USAGE_SKILLS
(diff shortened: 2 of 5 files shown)


wkentaro:

Reviewed the diff (`_collect_hunks` gains a required `usage` param, `USAGE_LIST` extracted and referenced from `HELP_LIST`, two e2e tests extended). Findings:

- **Correctness**: clean. Both call sites (`cmd_list` → `USAGE_LIST`, `cmd_show` → `USAGE_SHOW`) are updated and not swapped; `USAGE_LIST` renders byte-identically to the literal it replaced in `HELP_LIST`; `CliError(usage=...)` is a real kwarg that flows into the existing exit-2 path.
- **Consistency**: the extraction actually removes a pre-existing outlier — `HELP_LIST` was the only `HELP_*` block still inlining its `Usage:` line.
- **CI**: green across all 7 OS/Python matrix combos, which assert both the message and `Usage: git-hunk <cmd>` in stderr. Branch is current with `main` (0 commits behind), so no rebase.

Non-blocking nit (not worth a push): the new `usage` param is appended as the last keyword-only param, whereas sibling helpers `_apply_selection` / `_run_patch_command` put `usage` first. Cosmetic only; keyword-only so no call-site risk.

PR #120, opened 2026-07-13
test: cover commit rejecting an explicit blank -m message

wkentaro:
`commit` rejects a missing or blank `-m` message via
`if message is None or not message.strip()`, but the existing
`test_commit_requires_message` only omits `-m` entirely, so `message is None`
short-circuits and the `not message.strip()` half was never exercised. This adds
a parametrized case passing an explicit empty (`""`) and whitespace-only (`"   "`)
`-m` value, asserting the commit is refused with no change to the index.

## Test plan
- [x] `uv run pytest tests/e2e/commit_test.py` (10 passed)
- [x] `make lint` (ruff format/check + ty clean)

### DIFF
--- a/tests/e2e/commit_test.py
+++ b/tests/e2e/commit_test.py
@@ -1,5 +1,7 @@
 from pathlib import Path
 
+import pytest
+
 from .conftest import GitHunkCLI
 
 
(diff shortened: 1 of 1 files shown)


wkentaro:

Test-only PR adding a parametrized case for `git-hunk commit -m ""` / `-m "   "`. It covers a real gap: `test_commit_requires_message` only omits `-m`, hitting the `message is None` short-circuit at `_cli.py:606`, so the `not message.strip()` branch was never exercised. The new test drives that branch and asserts non-zero exit, `"message"` in stderr, and no commit/staging side effects.

- No rebase needed: branch is at merge-base with `main` (0 divergent commits, no file overlap).
- `/code-review` (high): clean, no findings. Uses existing helpers, matches file conventions.
- History: single, clear commit. `/verify` skipped (test-only, no runtime surface).
- CI: all 7 checks green (lint + test matrix across ubuntu/macos/windows × 3.10/3.14).

PR #121, opened 2026-07-14
refactor: raise on unparsable header in _extract_context_before

wkentaro:
The no-match branch in `_extract_context_before` is unreachable for real input: `parse_diff` only ever passes a real git `@@` hunk header, which always closes its range with a second `@@`. Converting the silent `return None` fallback into a `raise ValueError` fails loud if that invariant is ever broken, instead of silently dropping the section heading. Mirrors the fail-loud treatment PR #116 proposes for the sibling `_bare_header` helper.

## Test plan
- [x] `extract_context_before_test.py::test_raises_when_header_does_not_match` (updated from the old `is None` assertion)
- [x] `uv run pytest tests/` — 264 passed
- [x] `make lint` — ruff + ty clean

### DIFF
--- a/git_hunk/_hunk.py
+++ b/git_hunk/_hunk.py
@@ -182,7 +182,10 @@ def _bare_header(at_line: str) -> str:
 def _extract_context_before(header: str) -> str | None:
     match = re.search(r"@@.*?@@\s*(.*)", header)
     if not match:
-        return None
+        # Unreachable for real input: parse_diff only ever passes a real git @@
+        # hunk header, which always closes its range with a second @@. Raise
+        # rather than silently drop the heading if that invariant is ever broken.
+        raise ValueError(f"cannot parse hunk header: {header}")
     return match.group(1).strip() or None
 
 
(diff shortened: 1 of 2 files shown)


wkentaro:

**Verdict: recommend-merge**

Clean internal refactor. `_extract_context_before` turns its unreachable `return None` fallback into a `raise ValueError`, hardening the parse-diff invariant rather than silently dropping a hunk heading if it is ever broken.

Verified:
- The fallback is genuinely unreachable through the sole caller (`_hunk.py:324`). `parse_diff` only ever sees `git diff -U3` unified diffs (never combined `--cc`), `re.split(r"(?=^@@)", ...)` guarantees every `header_line` begins with `@@`, and a real git hunk header always closes its range with a second `@@` that the regex requires.
- Return type `str | None` is preserved (None still returned for whitespace-only trailing context via `.strip() or None`).
- The test is updated to assert the raise (`pytest.raises(ValueError, match=...)`); the `pytest` import is added and used.
- CI green across all 7 jobs; branch is `CLEAN` and sits directly on the current `main` tip (no rebase needed).

No CHANGELOG entry: the change is internal with no user-facing behavior difference.

PR #122, opened 2026-07-14
fix: surface a missing git executable as a clean CLI error

wkentaro:
When `git` is not on `PATH`, every `git-hunk` command crashed with a raw
`FileNotFoundError` traceback: `run_git` shelled out via
`subprocess.run(["git", ...])` with no handling for the binary being absent, so
the failure escaped before `_require_git_repo` could report its own styled
error. This converts that case into the tool's normal `error:` message.

`run_git` now translates the `FileNotFoundError` into the same `RuntimeError`
the git layer already raises for git failures, and `_require_git_repo` wraps its
`is_git_repo()` probe to surface it as a `CliError` (matching the existing
`RuntimeError -> CliError` translation used at the other call sites). Because
`is_git_repo()` is the first git call in every command, this covers all
subcommands.

Verified by driving the real CLI with `git` removed from `PATH`:

```
$ PATH=<empty> python -m git_hunk list
error: git executable not found on PATH   (exit 1)
```

`stage`, `show`, `commit`, and `list --json` all print the same clean error.

Scope note (deliberate): this fixes the missing-binary case only. A different
git failure (e.g. an unreadable `.git/index`) can still reach `get_diff` /
`get_untracked_files` on the `list`/`show`/`commit` paths, which are not wrapped
in `RuntimeError -> CliError`, so those surface as raw tracebacks today. Reviews
suggested centralizing the translation in `CliGroup.invoke` to close that class
of gap and remove the per-call-site duplication; that is a broader
error-handling change and left as a follow-up rather than folded into this
surgical fix.

## Test plan
- [x] Added `tests/e2e/error_test.py::test_missing_git_binary_reports_clean_error` (fails with a raw traceback before the fix, green after)
- [x] `uv run pytest` (265 passed)
- [x] Drove the real CLI with `git` off `PATH`: clean `error:`, exit 1, no traceback

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -71,5 +71,7 @@ and this project adheres to
   into a trailing subcommand, so `git-hunk -h stage src/foo.py` prints help
   without staging and `git-hunk -V list` prints the version without listing
   (#145).
+- Report a clean `error:` message instead of a raw `FileNotFoundError`
+  traceback when the `git` executable is not found on `PATH` (#122).
 
 [unreleased]: https://github.com/wkentaro/git-hunk/compare/v0.2.0...HEAD
--- a/git_hunk/_cli.py
+++ b/git_hunk/_cli.py
@@ -90,8 +90,11 @@ class CliGroup(click.Group):
 
 
 def _require_git_repo() -> None:
-    if not is_git_repo():
-        raise CliError("not a git repository")
+    try:
+        if not is_git_repo():
+            raise CliError("not a git repository")
+    except RuntimeError as exc:
+        raise CliError(str(exc)) from exc
 
 
 def _echo_json(data: object) -> None:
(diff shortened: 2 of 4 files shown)


wkentaro:

Full-finalize pass complete — recommending merge.

- **Rebase:** skipped. `main` gained 0 commits since the merge-base and merge state is CLEAN, so the existing green CI already reflects the merged result.
- **Review:** medium-effort `/code-review` over the diff found no meaningful issues. The fix is placed at the single `run_git` chokepoint (`FileNotFoundError` → `RuntimeError`), so every existing `RuntimeError`→`CliError` wrapper now also surfaces missing-git cleanly; the new `_require_git_repo` catch covers the `is_git_repo()` guard path. Deep fix, not a special-case.
- **Commits:** already a single atomic commit (fix + test + changelog) with a clean message — no reshaping needed.
- **Verify:** smoked the missing-git path with `git` absent from PATH → `error: git executable not found on PATH`, exit 1, no traceback.
- **CI:** all 7 checks green (lint + test across ubuntu/macos/windows × 3.10/3.14).

PR #124, opened 2026-07-14
fix: surface a git failure in commit's staged-change check as a clean CLI error

wkentaro:
`cmd_commit`'s already-staged guard called `get_diff(staged=True)` without wrapping the `RuntimeError` that `run_git` raises on a non-zero git exit. A git failure there (e.g. a corrupt `.git/index`, which still lets `git rev-parse --is-inside-work-tree` succeed) therefore crashed with a raw Python traceback instead of git-hunk's normal `error:` message. This wraps that one call in the same `RuntimeError -> CliError` idiom already used elsewhere in the file.

Scoped deliberately to the commit-specific `get_diff(staged=True)` call. The shared `_get_hunks` `get_diff` (used by `list`/`show`/`stage`/`unstage`/`discard`) has the same unguarded pattern, but that is the call site #107 already targets, so it is left to that PR. A broader fix (centralizing `RuntimeError -> CliError` at the top-level CLI dispatch, which would make several per-site wraps redundant) is worth a maintainer decision but is out of scope here.

## Test plan
- [x] `test_commit_reports_clean_error_when_staged_diff_fails` (corrupts `.git/index`, asserts a clean `error:` line and non-zero exit); confirmed it fails with a raw traceback on the unfixed code
- [x] drove the real `git-hunk commit` against a corrupt index -> `error: git diff -U3 --cached -- failed: fatal: .git/index: index file smaller than expected`, exit 1
- [x] `make lint` and full `pytest` green (265 passed)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -107,5 +107,8 @@ and this project adheres to
 - Report untracked files with repo-root-relative paths, matching tracked hunks,
   so running `git-hunk` from a subdirectory no longer emits an untracked `file`
   on a different path basis than the staged/unstaged hunks beside it (#103).
+- Report a git failure while checking for already-staged changes in `commit`
+  (e.g. a corrupt index) as a clean `error:` message instead of crashing with a
+  raw Python traceback (#124).
 
 [unreleased]: https://github.com/wkentaro/git-hunk/compare/v0.2.0...HEAD
--- a/git_hunk/_cli.py
+++ b/git_hunk/_cli.py
@@ -637,7 +637,11 @@ def cmd_commit(
         raise CliError("commit requires a message (-m)", usage=USAGE_COMMIT)
 
     _require_git_repo()
-    if get_diff(staged=True).strip():
+    try:
+        already_staged = get_diff(staged=True).strip()
+    except RuntimeError as exc:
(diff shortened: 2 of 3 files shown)


wkentaro:

**Verdict: recommend-merge**

Finalize pass:
- **Rebase:** skipped. `main` gained 0 commits since the merge-base and mergeState is `CLEAN`, so the green CI already reflects the merged result.
- **Review:** the change wraps `get_diff(staged=True)` in the same `except RuntimeError as exc: raise CliError(str(exc)) from exc` idiom already used at `_cli.py:276` and `:659` (the `_get_hunks` site from #107). Consistent with the established ad hoc translation pattern; no meaningful findings, so `/review-fix` was skipped.
- **Commit history:** a single well-scoped commit (src + test + CHANGELOG). No reshape needed.
- **Verify:** smoke-tested locally against a corrupt `.git/index` — `git-hunk commit` now surfaces `error: git diff ... failed: fatal: .git/index: index file smaller than expected` and exits 1 instead of a raw traceback. The bundled e2e test `test_commit_reports_clean_error_when_staged_diff_fails` covers the same path.
- **CI:** green across all 7 lint/test jobs (ubuntu/macos/windows × 3.10/3.14).

PR #126, opened 2026-07-15
test: cover the type-change merge path guard for a delete + unrelated add

wkentaro:
`parse_diff` merges a delete block immediately followed by an add block for the
*same* path into one synthetic type-change (`"T"`) hunk. The same-path check
(`extract_file_path(next_diff) == filepath`, `_hunk.py:262`) is what stops an
unrelated delete + add pair from collapsing into a bogus `"T"` hunk, but its
false branch (`262->275`) had no test — every existing type-change test uses the
same path on both blocks.

This adds one unit test: a delete of `a.txt` followed by an add of `b.txt` (git
sorts diff blocks by path, so this ordering is real) must stay two hunks (`D` and
`A`), not merge. Removing the guard collapses it into a single `T` hunk, so the
test kills a real regression rather than asserting the status quo.

## Test plan
- [x] `make test` (265 passed; the new test closes branch `_hunk.py 262->275`)
- [x] `make lint`

### DIFF
--- a/tests/unit/_hunk/parse_diff_test.py
+++ b/tests/unit/_hunk/parse_diff_test.py
@@ -208,3 +208,34 @@ def test_typechange_from_binary_is_marked_binary() -> None:
     assert hunks[0].binary is True
     assert hunks[0].a_mode == "100644"
     assert hunks[0].b_mode == "120000"
+
+
+def test_delete_then_unrelated_add_is_not_merged_as_typechange() -> None:
+    # The type-change merge only fires when the add block shares the delete's
+    # path. A delete followed by an add for a *different* file (git sorts diff
+    # blocks by path, so this is a real ordering) must stay two hunks, not merge
+    # into a bogus "T".
+    diff = (
+        "diff --git a/a.txt b/a.txt\n"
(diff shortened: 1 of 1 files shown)


wkentaro:

Test-only PR adding one focused case to `parse_diff_test.py`. It covers the false branch of the type-change merge guard at `git_hunk/_hunk.py:262` (`extract_file_path(next_diff) == filepath`): a `deleted`-block for `a.txt` immediately followed by a `new file` block for the *unrelated* `b.txt` must stay two hunks (`D` + `A`), not collapse into a bogus `T`. Since git sorts diff blocks by path, delete-then-unrelated-add is a real ordering, so this is a reachable guard, not a synthetic one.

- Assertions match the implementation; the test pins hunk count, file, `change_kind`, and add/delete counts.
- Divergence check: main gained 0 commits since the merge-base, so the branch is current and no rebase was needed.
- CI green across all 7 checks (lint + 6 test matrix legs).
- `/code-review` (medium) over the diff surfaced no meaningful findings; single clean commit, conventional message.

Ready for maintainer merge.

PR #128, opened 2026-07-15
test: cover a chmod + content edit in one file diff

wkentaro:
`git diff` emits a chmod plus a content edit to the same file as a single diff
block carrying both `old mode`/`new mode` headers and an `@@` body. Only the
mode-only case (`test_mode_only_change_surfaced`, no `@@` body) was covered, so
the branch where mode metadata rides along with a real content hunk was untested.
This pins that the content hunk carries `a_mode`/`b_mode` and the parsed diff,
rather than collapsing to the whole-file mode-only path.

## Test plan
- [x] `uv run pytest tests/unit/_hunk/parse_diff_test.py -q` (new `test_mode_and_content_change_together` passes; full suite 265 passed)
- [x] `uv run ruff check` / `uv run ty check` clean
- [x] fixture verified against real `git diff -U3` output for a `chmod +x` + edit

### DIFF
--- a/tests/unit/_hunk/parse_diff_test.py
+++ b/tests/unit/_hunk/parse_diff_test.py
@@ -126,6 +126,34 @@ def test_mode_only_change_surfaced() -> None:
     assert hunks[0].diff == ""
 
 
+def test_mode_and_content_change_together() -> None:
+    # A chmod plus an edit to the same file is one diff block carrying both the
+    # old/new mode headers and an @@ body. The content hunk must carry the mode
+    # metadata, not collapse to the whole-file mode-only path.
+    diff = (
+        "diff --git a/m.sh b/m.sh\n"
+        "old mode 100644\n"
+        "new mode 100755\n"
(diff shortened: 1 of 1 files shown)


wkentaro:

**Verdict: recommend-merge**

Test-only PR adding \`test_mode_and_content_change_together\`, which covers a chmod + content edit arriving in a single diff block, the gap between \`test_mode_only_change_surfaced\` (mode headers, no body) and the content-hunk tests. The assertions match real \`Hunk\` fields and the surrounding test conventions (single scenario, explanatory "why" comment).

- Branch is current with \`main\` (0 commits behind, merge state CLEAN), no rebase needed.
- \`/code-review\` (medium, report-only): no meaningful findings, so \`/review-fix\` skipped.
- Single, well-named commit, nothing to reshape.
- CI green across all 7 matrix jobs (lint + test on ubuntu/macos/windows x 3.10/3.14).

No push made this tick, so the existing green checks are authoritative.

PR #92, opened 2026-07-08
refactor: centralize --json output into a single _echo_json helper

wkentaro:
The three JSON-emitting commands (`list`, `show`, and the `skills` list/get/path subcommands) each spelled out `click.echo(json.dumps(x, indent=2))` — five identical sites. This folds them into one `_echo_json` helper so the pretty-print formatting lives in a single place.

Pure refactor: output is byte-identical, so the existing `--json` e2e tests (which assert exact output) cover it unchanged.

## Test plan
- [x] `uv run pytest -q tests/` — 256 passed
- [x] `uv run ruff check` / `uv run ty check` — clean
- [x] Smoked `list --json`, `show --json`, `skills list/path --json` — output unchanged

PR #129, opened 2026-07-15
refactor: dedup the JSON hunk envelope into a shared _echo_hunks_json helper

wkentaro:
`cmd_list` and `cmd_show` each built the identical `{"schema_version", "hunks"}`
JSON envelope, differing only in whether `to_dict` was passed `include_lines=True`.
Extract that into `_echo_hunks_json(hunks, *, include_lines=False)` next to
`_echo_json` so the envelope shape lives in one place. Follow-up to #92, which
centralized `_echo_json`.

Behavior-preserving: output bytes are identical, and the existing `list --json`
/ `show --json` e2e tests already exercise both call sites.

## Test plan
- [x] `uv run pytest` (264 passed)
- [x] smoked `git-hunk list --json` (no `lines`) and `show --json` (with `lines`) in a scratch repo

### DIFF
--- a/git_hunk/_cli.py
+++ b/git_hunk/_cli.py
@@ -104,6 +104,15 @@ def _echo_json(data: object) -> None:
     click.echo(json.dumps(data, indent=2))
 
 
+def _echo_hunks_json(hunks: list[Hunk], *, include_lines: bool = False) -> None:
+    _echo_json(
+        {
+            "schema_version": JSON_SCHEMA_VERSION,
+            "hunks": [h.to_dict(include_lines=include_lines) for h in hunks],
+        }
+    )
+
+
 def _get_hunks(staged: bool, files: list[str] | None = None) -> tuple[list[Hunk], str]:
     _require_git_repo()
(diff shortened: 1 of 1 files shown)


wkentaro:

**Verdict: recommend-merge**

Clean, behavior-preserving dedup. The identical JSON-envelope construction in `cmd_list` and `cmd_show` is extracted into `_echo_hunks_json(hunks, *, include_lines=False)`, with `include_lines` threading `show`'s line-level output through unchanged. The remaining `_echo_json` call sites (`skills list/get/path`) carry different payload shapes and are correctly left untouched.

- No rebase needed: branch sits directly on current `main` (0 commits gained since merge-base), so the green CI reflects the merged result.
- CI green across the full matrix (lint + test on ubuntu/macos/windows x 3.10/3.14); the `list --json` / `show --json` e2e tests exercise this exact rendering path.
- Internal refactor, not user-facing, so no CHANGELOG entry is warranted.

No follow-ups.

PR #133, opened 2026-07-16
test: cover that list --staged/--unstaged exclude untracked files

wkentaro:
`_collect_hunks` only adds untracked entries on the default (both-sides) path;
the `--staged` and `--unstaged` early-return branches deliberately exclude them.
That contract was documented but untested: none of the existing staged/unstaged
tests placed an untracked file in the repo, so a regression that leaked untracked
entries into a filtered listing would have gone undetected.

This adds `test_status_filters_exclude_untracked`, which puts a staged edit, an
unstaged edit, and a separate untracked file in the repo, then asserts each
filter returns exactly the one hunk of its own status.

## Test plan
- [x] `uv run pytest -q` (265 passed)
- [x] `uv run ruff check .`, `uv run ty check`
- [x] Mutated `_collect_hunks` to leak untracked into `--staged`; the new test fails as expected, then reverted

### DIFF
--- a/tests/e2e/list_test.py
+++ b/tests/e2e/list_test.py
@@ -189,6 +189,25 @@ def test_list_default_shows_untracked(cli: GitHunkCLI) -> None:
     assert untracked[0]["b_mode"] == "100644"
 
 
+def test_status_filters_exclude_untracked(cli: GitHunkCLI) -> None:
+    cli.repo.write_file("f.py", "old\n")
+    cli.repo.git("add", ".")
+    cli.repo.git("commit", "-m", "init")
+
+    cli.repo.write_file("f.py", "staged\n")
+    cli.repo.git("add", "f.py")
+    cli.repo.write_file("f.py", "unstaged\n")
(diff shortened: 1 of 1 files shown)


wkentaro:

**Verdict: \`recommend-merge\`**

Own, test-only PR adding e2e coverage that \`git-hunk list --staged\`/\`--unstaged\` exclude untracked files.

- No rebase needed: branch is already current with \`main\` (0 commits gained since merge-base, no file overlap), merge state \`CLEAN\`.
- \`/code-review\` (medium): clean. Helper signature (\`run_list_json(*args)\`) and call form match existing tests; assertions are non-vacuous (staged/unstaged each isolate one hunk, untracked excluded in both). Skipped \`/review-fix\`.
- \`/recommit\`: single clean commit, nothing to reshape.
- \`/verify\`: skipped (test-only, no runtime behavior).
- CI: all 7 checks green; no push this tick, so existing checks are authoritative.

PR #136, opened 2026-07-16
refactor: dedup hunk body-splitting into split_diff_body helper

wkentaro:
> *This was generated by AI.*

Removes the thin forwarding helpers while keeping one shared hunk body operation. `split_diff_body` now owns both header removal and trailing empty-line removal.

## Test plan
- [x] `make test` (324 passed)
- [x] `make lint`

### DIFF
--- a/git_hunk/_hunk.py
+++ b/git_hunk/_hunk.py
@@ -72,12 +72,8 @@ def is_whole_file_hunk(hunk: Hunk) -> bool:
 
 
 def _body_lines(diff: str) -> list[dict[str, Any]]:
-    if not diff:
-        return []
-    _, *body = diff.split("\n")
-    body = strip_trailing_empty_lines(body)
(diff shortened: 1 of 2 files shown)


wkentaro:

Pure behavior-preserving dedup: the three copies of "split a hunk body after the `@@` header, then drop trailing blank lines" now route through one `split_diff_body` helper.

- Empty-diff behavior is preserved: `split_diff_body("")` → `"".split("\n")[1:]` → `[]`, matching the old `if not diff: return []` guard.
- `parse_diff` still derives `header_line` from `part.split("\n", 1)[0]`, identical to the old unpack.
- The `_lines.py` import swap leaves no orphan `strip_trailing_empty_lines` had exactly one use there; `_hunk.py` still defines and uses it inside the new helper.

Single clean conventional commit, `mergeStateStatus: CLEAN`, all 7 CI checks green.

PR #138, opened 2026-07-17
test: cover staging a text and binary hunk together in one command

wkentaro:
git-hunk's pitch is grouping hunks by intent, so a single `stage`/`unstage` can
mix a text hunk (applied via a patch) with a whole-file binary hunk (staged
whole). `_apply_selection` splits the selection into `text` and `whole_file`
groups and applies both in one call, but every existing binary/mode/typechange
test exercised a single whole-file hunk alone. This covers the combined path.

## Test plan
- [x] `uv run pytest tests/e2e/binary_test.py -q` (9 passed)
- [x] `uv run ruff check` / `ruff format --check` / `ty check` clean

### DIFF
--- a/tests/e2e/binary_test.py
+++ b/tests/e2e/binary_test.py
@@ -100,6 +100,34 @@ def test_stage_added_binary(cli: GitHunkCLI) -> None:
     assert cli.repo.git("diff", "--cached").strip() == ""
 
 
+def test_stage_and_unstage_text_and_binary_hunk_together(cli: GitHunkCLI) -> None:
+    # git-hunk's pitch is grouping hunks by intent, so a single command may mix a
+    # text hunk (applied via a patch) with a whole-file binary hunk (staged whole).
+    root = Path(cli.repo.path)
+    (root / "t.txt").write_text("hello\n")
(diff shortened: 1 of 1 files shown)


wkentaro:

Test-only PR adding one e2e case (`tests/e2e/binary_test.py`) that covers staging a text hunk and a whole-file binary hunk together in one `stage` command, then unstaging both together — exercising git-hunk's core "group hunks by intent" path where a patch-applied text hunk and a whole-file binary hunk mix in a single invocation.

- Up to date with `main` (0 commits behind, no file overlap) — no rebase needed.
- Cheap `/code-review` pass: no meaningful findings. Uses existing fixtures (`cli`, `run_list_json`, `run_ok`); assertions are precise; the leading comment explains the intent rather than the mechanics. No `/review-fix` needed.
- History is a single clean, well-scoped commit — no `/recommit` reshape needed.
- Test-only, no runtime surface — `/verify` skipped.
- All 7 CI checks green (lint + test across ubuntu/macos/windows x 3.10/3.14); no push this tick, so the existing green is authoritative.

Sound and endorsed; merge is yours to make.

PR #145, opened 2026-07-18
fix: short-circuit global -h/-V so they don't run a trailing subcommand

wkentaro:
The top-level `cli()` group callback printed help/version and then `return`ed,
so Click (with `invoke_without_command=True`) went on to resolve and invoke the
trailing subcommand anyway. As a result `git-hunk -h stage src/foo.py` printed
the help text and then **actually staged the file**, and `git-hunk -V list`
printed the version and then ran `list`. The same fall-through applied to the
destructive `discard`/`commit` subcommands.

Calling `ctx.exit()` (instead of `return`) after printing makes the global
`-h`/`-V` flags short-circuit before any subcommand runs. `ctx.exit()` raises
`click.exceptions.Exit`, which isn't caught by `CliGroup.invoke`'s
`CliError`/`KeyboardInterrupt` handlers, so it propagates to Click's top-level
`main()` and exits 0 without dispatching the subcommand.

## Test plan
- [x] `tests/e2e/error_test.py::test_help_short_circuits_subcommand` — `-h stage <file>` prints help and does not stage
- [x] `tests/e2e/error_test.py::test_version_short_circuits_subcommand` — `-V list` prints version and does not list
- [x] `uv run pytest -q` (266 passed), `uv run ruff check .`, `uv run ruff format --check .`, `uv run ty check`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -59,5 +59,9 @@ and this project adheres to
   surfaces an untracked `foo.py` instead of silently dropping it, and path
   arguments resolve consistently across `list`/`stage`/`unstage`/`discard`
   regardless of the platform's path separator (#95).
+- Stop the global `-h`/`--help` and `-V`/`--version` flags from falling through
+  into a trailing subcommand, so `git-hunk -h stage src/foo.py` prints help
+  without staging and `git-hunk -V list` prints the version without listing
+  (#145).
 
 [unreleased]: https://github.com/wkentaro/git-hunk/compare/v0.2.0...HEAD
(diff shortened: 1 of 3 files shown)


wkentaro:

**Verdict: recommend-merge**

Finalized. The fix is correct and minimal: the group callback used `return`, which does not stop Click from invoking a trailing subcommand, so `git-hunk -h stage foo` printed help *and* staged, and `git-hunk -V list` printed the version *and* listed. Switching to `ctx.exit()` for both the version and help paths raises `SystemExit(0)`, which passes cleanly through `CliGroup.invoke` (it only catches `CliError`/`KeyboardInterrupt`) up to Click's standalone mode.

Reviewed:
- `/code-review` (medium): clean, no meaningful findings.
- History: single atomic commit (CHANGELOG + source + test), no reshape needed.
- Smoke: `-h stage f.py` prints help, exits 0, stages nothing; `-V list` prints version, exits 0, does not list.
- CI: 7/7 green (lint + test on ubuntu/macos/windows x 3.10/3.14). No push this tick, so the existing green reflects the merged result.

No follow-ups.

PR #147, opened 2026-07-19
docs: include type changes in the whole-file-hunk definition

wkentaro:
`CONTEXT.md` and ADR 0001 §5 defined a whole-file hunk as only a binary or
mode-only change, but `parse_diff` also emits a **type change** (`change_kind`
`"T"`, e.g. file ↔ symlink) as a whole-file hunk via the same `_whole_file_hunk`
constructor (`header: null`, empty body). `README.md` already lists all three
cases; this aligns the domain docs (which agents are told to treat as
authoritative) with the code and README.

## Test plan
- [x] Cross-checked against `git_hunk/_hunk.py` `_whole_file_hunk` call sites (type/binary/mode-only all constructed identically) and the existing `test_typechange_stage_unstage_discard` behavioral witness

### DIFF
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -4,11 +4,12 @@
 
 - **Hunk** — the atomic, addressable unit of git-hunk. A contiguous change a user can
   `stage` / `show` / `discard` by `id`. Usually one `@@` section of a unified diff; for
-  binary or mode-only changes it is a synthetic **whole-file hunk** (no `@@` range). The
-  top-level object in `--json` (the tool is hunk-centric, not file-centric).
+  binary, mode-only, or type changes it is a synthetic **whole-file hunk** (no `@@` range).
+  The top-level object in `--json` (the tool is hunk-centric, not file-centric).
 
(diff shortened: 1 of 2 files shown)


wkentaro:

**Verdict: recommend-merge**

Docs-only change aligning `CONTEXT.md` and ADR 0001 §5 with the code and README: a **type change** (`change_kind` `"T"`, e.g. file ↔ symlink) is also emitted as a whole-file hunk, alongside binary and mode-only changes.

Verified:
- `git_hunk/_hunk.py` `parse_diff` constructs a `_whole_file_hunk(..., change_kind="T", ...)` for typechanges (same constructor as binary/mode-only, `header: null`, empty body), so the previous docs undercounted the whole-file-hunk cases.
- `README.md` already listed all three cases; this closes the doc drift.
- Branch is up-to-date with `main` (no rebase needed); all 7 CI checks green.

No code/behavior change, so no `/verify`. Single clean, well-named commit. Ready for your review and merge.

PR #149, opened 2026-07-20
docs: correct stale-id claim for partial-stage leftovers

wkentaro:
The bundled agent skill guide, the README, and `_body_id`'s docstring/comment all
claimed a hunk's id is stable "across partial staging" and that a partially staged
hunk's leftover "keeps the same ID." That conflates two different cases:

- Staging *other* hunks leaves this hunk's id alone -- the id strips the `@@` line
  numbers that shift, so its body-hash is unchanged. This is the true claim.
- Staging only *part* of a hunk (via `-l`, or `--include-matching` /
  `--exclude-matching`) changes the leftover's body, so its content-hashed id
  changes. This is what the old wording got wrong.

An agent following the doc literally would reuse the original id on the leftover
and hit `error: hunk '<id>' not found`. Verified against the live CLI: a hunk
`85321a9`, after `stage 85321a9 -l 1`, leaves the leftover under a new id
`f130bce`, not `85321a9`.

This rewords every claim to say a partial stage gives the leftover a new id
(method-agnostic phrasing, since `--include-matching` behaves the same), and
points agents at `git-hunk list` to fetch it. The `_body_id` docstring and the
`_with_stable_ids` Pass-1 comment are corrected to match, so a contributor reading
the source doesn't reintroduce the claim.

**Deferred (conscious, not an oversight):** the two SKILL.md workflow examples
("Surgically drop debug lines" and "Separate a refactor from a feature") also reuse
a stale id after a partial stage and remain broken. "Surgically drop debug lines"
overlaps in-flight #109 on the exact `discard` line it edits, and fixing both needs
an example rewrite better done alongside #109, so they are left for a follow-up.

## Test plan
- [x] `make lint` and full suite (264 passed) green; docs/docstring-only, no behavior change
- [x] Corrected claim verified empirically against the `git-hunk` CLI (leftover gets a new id)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -97,5 +97,8 @@ and this project adheres to
   restoring "from HEAD" when it restores from the index: discarding an unstaged
   hunk reverts it to the staged content, leaving a staged sibling edit intact
   (#109).
+- Correct the skill guide and README's claim that a partially staged hunk's
+  leftover keeps the same id, so the guide now says to re-run `git-hunk list` to
+  get the leftover's new id (#149).
 
 [unreleased]: https://github.com/wkentaro/git-hunk/compare/v0.2.0...HEAD
--- a/README.md
+++ b/README.md
@@ -218,8 +218,9 @@ renaming, removing, or changing the type of an existing field bumps it. (Before
 3. For staging: reconstructs a minimal patch and pipes it through `git apply --cached`
 4. For discarding: reconstructs a reverse patch and applies it to the working tree
 
-IDs are stable across partial staging -- they are derived from the changed lines,
-not the `@@` line numbers that shift as you stage hunks.
+IDs are derived from the changed lines, not the `@@` line numbers that shift as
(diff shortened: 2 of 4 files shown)


wkentaro:

**Verdict: recommend-merge**

Docs + comment-only correction of an inaccurate ID-stability claim. Verified the corrected text is technically right: hunk IDs come from `_body_id`, a hash of the diff body (excluding shifting `@@` headers), so an unrelated hunk keeps its id while a partial stage rewrites the target hunk's body and therefore mints a **new** id for the leftover. The old wording ("the rest of the hunk stays ... with the same ID") was wrong; the fix is accurate and consistent across README, `SKILL.md`, the `_body_id`/`_with_stable_ids` docstring/comments, and the CHANGELOG entry (#149).

No runtime behavior changes, so no smoke needed. CI is green across all lint + test (ubuntu/macos/windows x 3.10/3.14) matrices, and the branch is `CLEAN` against `main` with zero divergence, so no rebase. Single, cleanly-messaged commit; no history reshaping needed.

PR #150, opened 2026-07-20
fix: match hunk ids case-insensitively

wkentaro:
An uppercased hunk id failed instead of resolving. Hunk ids are always
lowercase sha256 prefixes, but the CLI matched them case-sensitively, so
`git-hunk show 713B7B9` errored with a misleading "not found" and
`git-hunk stage 713B7B9` fell through to "no changed file matches" (the
uppercase arg failed the `[0-9a-f]+` shape check and was treated as a missing
path). git's own object-id lookup is case-insensitive (`git show DEADBEEF`
works), so this now mirrors that.

`_find_hunks_by_ids` lowercases the query before matching (used by `show`), and
`_select_hunks`' hex shape-check regex now accepts `A-F` so uppercase args route
to id-matching (used by stage/unstage/discard/commit). Error messages still echo
the user's original-case input.

## Test plan
- [x] unit: `_find_hunks_by_ids` resolves `AB12` / `Ab12` / `AB12CD0`
- [x] e2e: `show`/`stage` by uppercased id
- [x] `uv run pytest -q` (269 passed), `ruff check`, `ty check`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -81,5 +81,7 @@ and this project adheres to
   (#107).
 - Show a `Usage:` hint on the `--staged`/`--unstaged` conflict error for `list`
   and `show`, matching every other conflicting-flags error (#117).
+- Match hunk ids case-insensitively, so an uppercased id like `git-hunk show 713B7B9` resolves like git's own object-id lookup instead of failing with a
+  misleading "not found" / "no changed file matches" error (#150).
 
 [unreleased]: https://github.com/wkentaro/git-hunk/compare/v0.2.0...HEAD
--- a/git_hunk/_cli.py
+++ b/git_hunk/_cli.py
@@ -120,7 +120,7 @@ def _find_hunks_by_ids(hunks: list[Hunk], ids: list[str]) -> list[Hunk]:
     for hunk_id in ids:
         if not hunk_id.strip():
             raise CliError("hunk id must not be empty or whitespace")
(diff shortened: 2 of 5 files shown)


wkentaro:

CI was red on `lint` because the new CHANGELOG entry wasn't `mdformat`-clean (a soft line break inside the ```git-hunk show 713B7B9``` inline-code span). Ran `make format`, folded the fix into the single commit, and force-pushed. Verified locally: full `make lint` green, affected tests pass, and an uppercased id (`4D42FE1`) now resolves to the same hunk as its lowercase form. Awaiting CI re-run before a verdict.

wkentaro:

CI is green across all matrix jobs (lint + test on ubuntu/macos/windows × 3.10/3.14) and the branch is already up to date with `main` (0 commits behind the merge-base), so no rebase was needed this tick.

Reviewed the diff: the two id-matching sites are both covered, `_find_hunks_by_ids` lowercases the query before the `startswith` match, and `_select_hunks` widens the hex shape-check to `[0-9a-fA-F]` so uppercase args route to id-matching instead of failing the path check. No other case-sensitive id matcher or `[0-9a-f]` shape check exists in `_cli.py`, so the fix is complete. Error messages still echo the original-case input.

Smoke-verified end-to-end against the branch on a scratch repo: `git-hunk show FD675C0` and `git-hunk stage FD675C0` (uppercased id) both resolve the same hunk their lowercase forms do.

Verdict: **recommend-merge**.

PR #155, opened 2026-07-22
fix: scope the single-hunk help note to all line-selection options

wkentaro:
The `-l`, `--include-matching`, and `--exclude-matching` flags all share the
same single-hunk constraint (`_apply_line_filter` rejects any active selection
when more than one hunk is targeted), but `--help` attached the
"(requires exactly one hunk)" note only to the `-l` row. Since this help block
is shared across `stage`, `unstage`, `discard`, and `commit`, the omission read,
on four screens, as if pattern-based selection worked across multiple hunks.

Move the note to a group-level line covering all three options, placed after
`--exclude-matching` and before `--regex` (which is validated separately and
does not carry the constraint). Also add an e2e test asserting
`--include-matching` is rejected on a multi-hunk path selection; previously only
`-l` was covered.

## Test plan
- [x] `tests/e2e/path_shorthand_test.py::test_matching_selection_rejected_for_multi_hunk_path`
- [x] Rendered `git-hunk {stage,unstage,discard,commit} --help`; all four show the group-scoped note
- [x] `uv run pytest -q` (265 passed), `ruff check`, `ruff format --check`, `ty check`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -110,5 +110,9 @@ and this project adheres to
 - Report a git failure while checking for already-staged changes in `commit`
   (e.g. a corrupt index) as a clean `error:` message instead of crashing with a
   raw Python traceback (#124).
+- Scope the "requires exactly one hunk" note in `--help` to all line-selection
+  options, so `--include-matching` / `--exclude-matching` no longer read as
+  working across multiple hunks when they share the same single-hunk
+  constraint as `-l` (#155).
 
 [unreleased]: https://github.com/wkentaro/git-hunk/compare/v0.2.0...HEAD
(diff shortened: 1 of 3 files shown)


wkentaro:

**Verdict: recommend-merge**

Reviewed the diff (help-text scoping + mirror test + changelog). Verified against `git_hunk/_cli.py`: `_apply_line_filter` guards on `_Selection.is_active()`, which is true for `-l`, `--include-matching`, and `--exclude-matching` alike, so the relocated group-level note ("Line selection ... requires exactly one hunk") is factually accurate and not a regression from the old `-l`-only wording. `--regex` is correctly left outside the note (it carries no hunk-count constraint). The new `test_matching_selection_rejected_for_multi_hunk_path` correctly mirrors the existing `-l` test. Docs/test-only, no runtime behavior change.

CI: all 7 checks green (lint + ubuntu/macos/windows × 3.10/3.14). No rebase needed (branch has no divergence from `main`).

PR #158, opened 2026-07-23
refactor: dedup the list/show status+json option block into a helper

wkentaro:
`cmd_list` and `cmd_show` carried a byte-for-byte identical stack of four
`click.option` decorators (`--staged`, `--unstaged`, `--json`, `-h/--help`).
Extract it into `_add_status_json_options` so the shared flag set lives in one
place. The options are applied in reversed order, which preserves the exact
click parameter order, so `--help` text and argument parsing are unchanged.

Scoped to `list`/`show` only. `stage`/`unstage`/`discard` share a *different*
five/six-option set (that duplication is a separate concern, cf. #106), and
`skills` shares only a subset, so neither is folded in here.

## Test plan
- [x] `uv run ruff check` + `uv run ty check` clean
- [x] verified click param order is identical for both commands after the change
- [x] smoke: `list`, `list --json`, `list --staged`, `show <id>`, `list -h` all render as before

### DIFF
--- a/git_hunk/_cli.py
+++ b/git_hunk/_cli.py
@@ -3,6 +3,7 @@ import os
 import posixpath
 import re
 import stat
+from collections.abc import Callable
 from dataclasses import dataclass
 from dataclasses import replace
 from typing import Final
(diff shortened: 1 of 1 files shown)


wkentaro:

Behavior-preserving dedup of the four shared `--staged/--unstaged/--json/-h` options for `list` and `show` into `_add_status_json_options`. Verified the helper preserves option registration order (decorators apply bottom-up, so original outermost `--staged` maps to the helper applying `staged` last) and keeps every flag name and dest (`force_json`, `show_help`) identical. `cmd_list`/`cmd_show` signatures are unchanged, so no call site is affected.

- Merge state: CLEAN, 0 commits behind base (no rebase needed).
- CI: all 7 checks green (lint + test matrix across ubuntu/macos/windows × 3.10/3.14).
- Review (medium, report-only): no findings.
- Single clean logical commit; no reshape needed.

Issue #22, opened 2026-06-09
Stage untracked/new files


wkentaro:
Audited on 2026-09-18 against [`5e46e0f`](https://github.com/wkentaro/git-hunk/tree/5e46e0f5dceb76a377dc06bda1421f7cd74ae688). This brief supersedes stale implementation assumptions in earlier comments; historical decisions remain linked below.

## What to build

Let git-hunk stage untracked/new files. Today untracked files appear in `list`
with an empty ID and cannot be acted on by any git-hunk command — users must
fall back to `git add`. This is the single most common real-world friction
(seen across many sessions, with `git add -N` discovered as a manual
workaround).

This needs a mechanism decision before implementing:
- Surface untracked files in `list` with real, stable hunk IDs (e.g. by treating
  a new file as one all-lines hunk, or by intent-adding under the hood), and make
  `stage` act on them.
- Decide the discard/unstage semantics for a never-committed file (and guard
  against destroying untracked content).

## Acceptance criteria

- [ ] An untracked file shows a real, actionable hunk ID in `list`/`list --json`.
- [ ] `git-hunk stage <id>` (or path, per #13) stages a new file's contents.
- [ ] Discard/unstage behavior for never-committed files is defined, safe, and
      documented (no silent data loss).
- [ ] e2e tests cover listing and staging an untracked file end-to-end.

## Blocked by

None - can start immediately.

## Audit update — 2026-09-18

Still absent at the audited revision: `_get_untracked_entries` creates unaddressable inventory entries, and mutation selects tracked staged/unstaged hunks. README and CONTEXT explicitly document the empty ID. The September usage evidence in the comments remains relevant. No implementation or open PR closes this capability gap.

Keep the original mechanism/safety decision, including never-committed-file discard/unstage behavior. Coordinate ordinary copied-file behavior with #200. The enumeration failures themselves are consolidated under #139 and should not widen this feature. v0.3's freeze is no longer a blocker; #227 owns current prioritization.

wkentaro:
> *This was generated by AI during a Claude/Codex log audit on 2026-09-04.*

## Fresh usage evidence

This remains the dominant genuine `git-hunk` workflow failure in recent agent logs. Nine Claude session logs from the last 14 days contain `error: no changed file matches ...` for newly created files, including towncrier fragments, new tests, a new `AGENTS.md`, and a new decision artifact. The agents then fall back to `git add` plus `git commit`, exactly the detour this issue is meant to remove.

[...]

PR #162, opened 2026-07-24
docs: correct that untracked entries carry no addressable hunk id

wkentaro:
`git-hunk list` groups changes under `staged`, `unstaged`, and `untracked`, but
only the first two are addressable. The docs did not say so:

- README's JSON field table described `id` as, unconditionally, a "stable,
  content-based hunk id ... accepts prefixes".
- SKILL.md's "Reading the output" introduced all three groups and then said
  "Each hunk line is `id`, the `@@` header with its enclosing context, then
  `+N -N`."

Untracked entries have none of that. `_get_untracked_entries` builds them with
`id=""`, `print_hunk_list` renders that group with `show_hunks=False` so only
bare paths print, and `stage`/`unstage`/`discard`/`commit` never load untracked
entries at all (they source hunks from `_get_hunks`, which is diff-based). So an
agent following the skill doc looks for an id that does not exist, then falls
back to the path and hits `error: no changed file matches '<path>'`. This
corrects both docs to describe the `untracked` group as inventory only and
points at `git add` as the way in.

Docs only -- no behavior change. Staging untracked files remains #22.

Deliberately out of scope: README's `header` row also lists only "(binary,
mode-only, or type)" as the causes of `null`, though untracked entries carry
`header: null` too. Adding "untracked" there would imply an untracked entry is a
whole-file hunk, which contradicts the `CONTEXT.md` glossary that #147 is
currently amending -- worth a separate look once #147 lands.

## Test plan
- [x] Verified all three claims against a scratch repo: plain `list` prints
      untracked entries as bare paths; `list --json` emits `"id": ""`; and
      `stage`/`unstage`/`discard`/`commit`/`show` all reject an untracked path
- [x] `make lint` (mdformat re-flowed the README table's column padding after
      the `id` cell grew)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -100,5 +100,9 @@ and this project adheres to
 - Correct the skill guide and README's claim that a partially staged hunk's
   leftover keeps the same id, so the guide now says to re-run `git-hunk list` to
   get the leftover's new id (#149).
+- Correct the README JSON table and skill docs, which described every listed
+  hunk as carrying a usable `id`: an `untracked` entry's `id` is empty, so no
+  git-hunk command can address it, and plain `list` renders it as a bare path
+  (#162).
 
 [unreleased]: https://github.com/wkentaro/git-hunk/compare/v0.2.0...HEAD
--- a/README.md
+++ b/README.md
@@ -168,22 +168,22 @@ body; `show --json` adds a structured `lines` array. A `show --json` hunk
 }
 ```
 
-| Field            | Type           | Description                                                                                                  |
-| ---------------- | -------------- | ------------------------------------------------------------------------------------------------------------ |
-| `schema_version` | int            | Envelope version; bumped on any incompatible change to the shape below.                                      |
-| `hunks`          | array          | The hunks (empty array when there are no changes).                                                           |
-| `id`             | string         | Stable, content-based hunk id (7-char SHA-256 prefix); accepts prefixes.                                     |
-| `file`           | union          | Path of the changed file, as a byte-safe `{text\|bytes}` union (see below).                                  |
-| `status`         | string         | One of `staged`, `unstaged`, `untracked`.                                                                    |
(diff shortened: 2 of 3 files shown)


wkentaro:

**Freshness.** Branch head `3d0e36a` sits directly on current `origin/main` (0 commits gained since the merge-base, no file overlap), merge state `CLEAN`. No rebase needed, so the existing checks are authoritative: all 7 green (lint + test across ubuntu/macos/windows x 3.10/3.14).

**Review.** A report-only `/code-review` pass over both axes came back clean.

- *Spec:* no missing requirements, no scope creep. Every claim in the PR description was verified against the code: `id=""` at `_cli.py:334`, `show_hunks=False` at `_ui.py:119`, and `stage`/`unstage`/`discard`/`commit` routing solely through `_get_hunks` so untracked entries never enter the addressable pool. The `header` row was correctly left untouched, matching the stated out-of-scope note.
- *Standards:* CHANGELOG entry correctly placed under `## [Unreleased]` with `(#162)` per `CLAUDE.md`; no em-dashes introduced; calling an id-less untracked item an "entry" rather than a "hunk" is consistent with the `CONTEXT.md` glossary, which defines a hunk as addressable by id.
- Only judgement-call nit: the clause "no git-hunk command can address it" appears in both README and `SKILL.md`. The two serve different audiences (JSON schema reference vs. CLI reading guide), so the duplication is justified; not worth a change.

**Independent smoke test.** Confirmed in a scratch repo that plain `list` prints untracked files as bare paths under `untracked:` while tracked hunks print an id, and `list --json` emits `"id": ""` for the untracked entry. The docs now match observed behaviour.

**Commits.** One commit, one concern, message explains the why. Nothing to reshape.

follow-ups: #163 (README's `header: null` row omits untracked entries as a fourth `null` cause; deliberately deferred by this PR pending #147, does not block this merge)

PR #164, opened 2026-07-25
test: cover that a multi-target selection applies nothing when one target is unknown

wkentaro:
`stage`, `unstage`, and `discard` resolve every target (hunk id or file path)
before applying anything, so one unknown target aborts the command with nothing
applied. Nothing asserted that. The existing error tests all pass a *single*
bad target, which cannot distinguish "aborted cleanly" from "applied what it
could, then failed" — so reordering validation after the first apply would
half-apply a selection and the suite would stay green.

Assertions name the specific unknown target rather than just matching `not
found`, since with two targets a generic match cannot tell a correct rejection
from the CLI flagging the wrong one.

## Test plan

- [x] 4 tests added: `tests/e2e/selection_atomicity_test.py`
- [x] mutation-checked — a resolve-and-apply-per-target mutant of
      `_apply_selection` fails all 4; a mutant naming a valid id in the error
      instead of the bogus one fails the 3 id tests
- [x] `make lint` and full suite green (268 passed)

### DIFF
--- a/tests/e2e/selection_atomicity_test.py
+++ b/tests/e2e/selection_atomicity_test.py
@@ -1,0 +1,77 @@
+"""A multi-target selection is resolved fully before anything is applied.
+
+Every id and path is validated up front, so one bad target aborts the whole
+command and the valid targets alongside it are left untouched. Without this,
+reordering validation after the first apply would silently half-apply a
+selection.
+"""
+
+import pytest
+
+from .conftest import GitHunkCLI
+
+
+@pytest.fixture
+def two_changed_files(cli: GitHunkCLI) -> GitHunkCLI:
+    cli.repo.write_file("f1.py", "a\n")
+    cli.repo.write_file("f2.py", "b\n")
+    cli.repo.git("add", ".")
+    cli.repo.git("commit", "-m", "init")
+    cli.repo.write_file("f1.py", "A\n")
+    cli.repo.write_file("f2.py", "B\n")
+    return cli
+
+
+def test_stage_with_one_unknown_id_stages_nothing(
+    two_changed_files: GitHunkCLI,
+) -> None:
+    cli = two_changed_files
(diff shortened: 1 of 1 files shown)


wkentaro:

**Rebase:** skipped. `origin/main` has gained 0 commits since the merge-base, so the branch already sits on current `main` and the green CI reflects the merged result exactly. Merge state is `CLEAN`.

**Review (`/code-review`, medium, report-only):** clean on both axes, so no `/review-fix` escalation.

- *Standards* — follows house convention: plain `test_` functions, shared arrange in the `two_changed_files` fixture, full signature annotations, aspect-named file. The stage/unstage/discard triplication mirrors the established shape in `path_shorthand_test.py`, so it is not a duplication smell. One judgement-call nit: this is the only test file in the repo carrying a module-level docstring; the content is legitimate "why" reasoning, so it stays a style call, not a violation.
- *Spec* — every claim in the description was verified against the code, including by applying both described mutants. A resolve-and-apply-per-target mutant of `_apply_selection` fails all 4 tests; a mutant naming a valid id in the error instead of the bogus one fails the 3 id tests. The gap claim holds: `error_test.py`, `dry_run_test.py`, `commit_test.py`, `markup_test.py`, `path_shorthand_test.py` and `find_hunks_by_ids_test.py` each pass exactly one bad target, so none can distinguish "aborted cleanly" from "half-applied". No scope creep; omitting a CHANGELOG entry matches how other test-only commits on `main` are handled.

**`/recommit`:** no-op. The branch is a single, atomic, conventionally-titled commit.

**`/verify`:** skipped. Test-only, no runtime behaviour change.

**CI:** all 7 checks green (lint, plus test on ubuntu/macos/windows x 3.10/3.14). Reproduced locally: `make lint` clean, `uv run pytest -q` 268 passed.

No follow-up issues filed.

PR #165, opened 2026-07-25
test: pin the CLI exit-code contract (2 usage, 1 domain)

wkentaro:
`CliGroup.invoke` exits 2 when a `CliError` carries a usage block and 1 when it
does not, so a scripted caller can tell "you invoked this wrong" from "what you
named does not resolve". Nothing outside `tests/e2e/skills_test.py` pinned that
split: every other error test asserted only `returncode != 0`, so collapsing the
branch to a single exit code would have passed CI.

The new test asserts both halves together, because the exit code and the presence
of the usage block are one decision in `_cli.py`. Each parametrized case reaches a
distinct `raise CliError(...)` site, so a raise site that silently loses its
`usage=` argument fails here too — which a unit test on the `ctx.exit(...)` line
alone would not catch.

Test-only; no production code and no user-facing behavior changes, so no CHANGELOG entry.

## Test plan
- [x] `make test` — 273 passed
- [x] `make lint` — clean
- [x] mutation check: replacing `ctx.exit(2 if exc.usage else 1)` with `ctx.exit(1)` fails 5 of the 9 new cases

### DIFF
--- a/tests/e2e/exit_code_test.py
+++ b/tests/e2e/exit_code_test.py
@@ -1,0 +1,66 @@
+import pytest
+
+from .conftest import GitHunkCLI
+
+
+@pytest.fixture
+def cli_with_change(cli: GitHunkCLI) -> GitHunkCLI:
+    cli.repo.write_file("f.py", "a\nb\nc\n")
+    cli.repo.git("add", ".")
+    cli.repo.git("commit", "-m", "init")
+    cli.repo.write_file("f.py", "A\nb\nC\n")
+    return cli
+
+
+# A caller that got the invocation wrong exits 2 and is shown how to invoke it;
+# a caller that invoked correctly but named something unresolvable exits 1. The
+# two halves are one decision in CliGroup.invoke, so they are pinned together.
+
+
+@pytest.mark.parametrize(
+    "args",
+    [
+        ("bogus",),
+        ("stage",),
+        ("stage", "--regex"),
+        ("stage", "-l", "1", "--include-matching", "x"),
+        ("commit", "f.py"),
+    ],
+    ids=[
+        "unknown-subcommand",
+        "stage-without-target",
+        "regex-without-pattern",
+        "conflicting-selectors",
+        "commit-without-message",
+    ],
+)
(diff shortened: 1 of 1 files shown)


wkentaro:

**Rebase:** not needed. `origin/main` has gained **0 commits** since the merge-base,
merge state is `CLEAN`, and the branch head equals its remote. Left untouched, so no
push this tick and the existing checks are authoritative.

**Review:** one report-only two-axis pass (Standards + Spec), no `--fix`. No
meaningful findings on either axis, so `/review-fix` was skipped.

- *Standards* — conforms to the repo conventions: no test classes, fully annotated
  signatures, explicit `ids=` on both `parametrize` blocks, no blank line between act
  and asserts, shared arrange deduplicated into the `cli_with_change` fixture, and
  the one comment explains *why* the two halves are pinned together rather than
  restating the code. Every parametrized case was independently verified to reach the
  branch its id claims.
- *Spec* — the test is a **true pin**, not a vacuous assertion, confirmed by mutation
  testing. Dropping `usage=` from the `commit` "requires a message" raise site made
  the `commit-without-message` case fail; changing the `CliGroup.invoke` dispatch to
  an unconditional `ctx.exit(1)` made all five usage cases fail while the four domain
  cases correctly stayed green. The nine cases map to nine **distinct** raise sites
  with no collapsing, so the parametrization is as broad as it looks.

This also confirms the contract has two halves that only an e2e test pins together:
the dispatch line alone does not encode *which* `raise CliError(...)` sites carry
`usage=`, so a raise site silently losing that argument is caught here and would not
be caught by a unit test on the dispatch.

**Commits:** already a single, logically self-contained commit with a message that
explains the gap it closes. `/recommit` was a no-op; nothing was reshaped or pushed.

**Smoke test:** skipped — test-only, no runtime behaviour change. No `CHANGELOG`
entry needed for the same reason.

**CI:** green on all 7 checks (lint, plus test on ubuntu/macos/windows x 3.10/3.14).

follow-ups: #166 (the remaining success-banner and empty-state lines the suite
renders but never asserts; does not block this merge)

PR #167, opened 2026-07-25
test: assert the success banners and empty-state lines the suite never checked

### LINKED ISSUE #166: Test: assert the success banners and empty-state lines the suite renders but never checks

Follow-up noticed while processing #165, which pinned the first of this family
(the CLI exit-code contract). It does not block that PR.

`_ui.py` renders several user-facing lines that the existing suite *executes* but
never *asserts*, so a regression in any of them stays green.

### 1. Real (non-dry-run) success banners are unasserted

`print_applied` (`git_hunk/_ui.py:175`) and `print_committed` (`git_hunk/_ui.py:191`)
produce the `staged <id> <file> ...` / `committed N hunks <subject>` banners on the
real path. The suite only asserts the `would ...` dry-run text; nothing reads the
real banner. Verified on `origin/main`:

```
$ git grep -n "Staged \|Committed \|Unstaged \|Discarded " origin/main -- tests/ | grep -v would
(no output)
```

A change to the verb, the id/file ordering, or the pluralisation in
`print_committed` (`f"{count} hunk{'s' if count != 1 else ''}"`) would not fail
any test.

### 2. Empty-state lines are unasserted

`No hunks.` (`git_hunk/_ui.py:106`) and `No skills.` (`git_hunk/_ui.py:156`) both go
to **stderr**. No test reads either string:

```
$ git grep -n "No hunks\.\|No skills\." origin/main -- tests/
(no output)
```

`tests/e2e/list_test.py::test_empty_new_file_is_not_a_bogus_mode_hunk` does execute
the `No hunks.` branch, but asserts only on stdout, so it passes vacuously — the
branch could print nothing at all and the test would still be green.

### 3. `show` excludes untracked entries by design, unasserted

`show` is invoked with `include_untracked=False` while `list` includes untracked
entries. That asymmetry is deliberate, but `tests/e2e/show_test.py` has no test
pinning it.

### Scope note

A fourth gap from the same sweep — that an untracked path `list` displays can never
be staged through git-hunk — is deliberately **not** included here, because pinning
that boundary overlaps #22 (`Stage untracked/new files`) and should follow that
issue's direction rather than cement the current behaviour.

### Suggested shape

Three small e2e tests (or additions to the existing `list_test.py` / `show_test.py`),
asserting on **stderr** for items 1 and 2. Each should be mutation-checked: break the
banner in `_ui.py` and confirm the new test actually fails.

wkentaro:
## Summary

Pins user-facing output contracts the suite already executed but never asserted, so a regression in any of them stayed green:

- The real (non-dry-run) `staged` / `unstaged` / `discarded` banners and the `committed N hunk[s] <subject>` banner, all on stderr. Only the `would ...` dry-run text was asserted before. The new tests pin the whole rendered line — indent, verb, hunk id, file, `@@` header, `+N -N` stats — one line per selected hunk, and the commit banner in the singular, plural, and multi-line-message (subject only) directions.
- The `No hunks.` and `No skills.` empty-state lines, both on stderr. `list_test.py::test_empty_new_file_is_not_a_bogus_mode_hunk` reached the `No hunks.` branch but asserted only on stdout, so it passed vacuously. Per the brief, the `No skills.` assertion extends the existing empty-skills-dir test rather than adding a duplicate scenario.
- That `show` omits untracked entries which `list` reports (`include_untracked=False` vs `True`). The precondition is checked through `list --json` carrying `"status": "untracked"`.

These are the suite's first verbatim full-line stderr assertions, and rich sizes its output from `COLUMNS`, so the banner lines wrapped and the tests failed under a narrow terminal. The first commit pins the e2e console width in `GitHunkCLI.run`, which keeps the exact assertions honest without weakening them.

No file under `git_hunk/` is modified, so there is no CHANGELOG entry (matching precedent for test-only changes).

Closes #166

## Mutation checks

Each assertion was mutation-checked: the corresponding string or branch was broken in the renderer, the affected test confirmed failing, then reverted. `git_hunk/` is unmodified at HEAD.

| # | Mutation | Result |
| --- | --- | --- |
| 1 | `print_applied`: swap hunk-id and file order | KILLED |
| 2 | `print_applied`: drop the `+N/-N` stats | KILLED |
| 3 | `print_applied`: uppercase the verb | KILLED |
| 4 | `print_committed`: drop the plural `s` | KILLED (pluralize test) |
| 5 | `print_committed`: rename `committed` to `commit` | KILLED (singular test) |
| 6 | `print_hunk_list`: silence `No hunks.` | KILLED |
| 7 | `print_skill_list`: silence `No skills.` | KILLED |
| 8 | `cmd_show`: `include_untracked=True` | KILLED |
| 9 | `print_applied`: report only the first hunk (`for hunk in hunks[:1]`) | KILLED (one-line-per-hunk test) |
| 10 | `print_committed`: use the whole message, not `splitlines()[0]` | KILLED (subject-only test) |
| 11 | `print_applied` / `print_committed`: drop the two-space indent | KILLED (all banner tests) |

Mutations 9 and 10 survived the whole suite before this round: nothing pinned the per-hunk banner loop or the subject extraction.

## Test plan

- [x] `uv run pytest -q` — 336 passed
- [x] `COLUMNS=40 uv run pytest -q` — 336 passed (3 banner tests failed here before the width pin)
- [x] `uv run ruff check .`, `uv run ruff format --check .`, `uv run ty check` — clean
- [x] Each of the four commits independently green (lint, format, typecheck, full suite) in a detached checkout
- [x] Rebased on current `main`

### DIFF
--- a/tests/e2e/conftest.py
+++ b/tests/e2e/conftest.py
@@ -16,7 +16,9 @@ class GitHunkCLI:
         self.repo = repo
 
     def run(self, *args: str) -> subprocess.CompletedProcess[str]:
-        runner = CliRunner()
+        # rich sizes its output from COLUMNS; pin it so the lines tests assert
+        # verbatim do not wrap when the suite runs in a narrow terminal.
+        runner = CliRunner(env={"COLUMNS": "200"})
         old_cwd = os.getcwd()
         try:
             os.chdir(self.repo.path)
--- a/tests/e2e/list_test.py
+++ b/tests/e2e/list_test.py
@@ -221,6 +221,16 @@ def test_empty_new_file_is_not_a_bogus_mode_hunk(cli: GitHunkCLI) -> None:
     assert "Mode None" not in cli.run_ok("list", "--staged")
 
 
+def test_no_hunks_message_goes_to_stderr(cli: GitHunkCLI) -> None:
+    cli.repo.write_file("f.py", "x\n")
+    cli.repo.git("add", ".")
+    cli.repo.git("commit", "-m", "init")
+
+    r = cli.run("list")
+    assert r.returncode == 0
+    assert "No hunks." in r.stderr
+
+
 def test_list_unstaged_filter(cli: GitHunkCLI) -> None:
     cli.repo.write_file("f.py", "old\n")
     cli.repo.git("add", ".")
--- a/tests/e2e/show_test.py
+++ b/tests/e2e/show_test.py
@@ -92,6 +92,23 @@ def test_show_no_args_shows_all(cli: GitHunkCLI) -> None:
     assert "+unstaged" in r.stdout
 
 
+def test_show_excludes_untracked_files_that_list_includes(cli: GitHunkCLI) -> None:
+    cli.repo.write_file("tracked.py", "old\n")
+    cli.repo.git("add", ".")
+    cli.repo.git("commit", "-m", "init")
+    cli.repo.write_file("tracked.py", "new\n")
+    cli.repo.write_file("fresh.py", "fresh\n")
+
+    listed = cli.run_list_json("list", "--json")
+    untracked = [h for h in listed if h["status"] == "untracked"]
+    assert [h["file"]["text"] for h in untracked] == ["fresh.py"]
+
+    r = cli.run("show")
+    assert r.returncode == 0
+    assert "tracked.py" in r.stdout
+    assert "fresh.py" not in r.stdout
+
+
 def test_show_staged_and_unstaged_together_errors(cli: GitHunkCLI) -> None:
     cli.repo.write_file("f.py", "old\n")
     cli.repo.git("add", ".")
--- a/tests/e2e/skills_test.py
+++ b/tests/e2e/skills_test.py
@@ -116,11 +116,13 @@ def test_list_rejects_arguments(cli: GitHunkCLI) -> None:
     assert "takes no arguments" in result.stderr
 
 
-def test_list_empty_skills_dir_exits_zero(
+def test_list_empty_skills_dir_exits_zero_and_says_no_skills(
     cli: GitHunkCLI, tmp_path: Path, monkeypatch: pytest.MonkeyPatch
 ) -> None:
     monkeypatch.setenv("GIT_HUNK_SKILLS_DIR", str(tmp_path))
-    assert cli.run("skills", "list").returncode == 0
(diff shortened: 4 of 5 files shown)


wkentaro:

**Rebase:** not needed. `origin/main` gained 0 commits since the merge-base (`3c087be`), merge state is `CLEAN`, so the existing green reflects the merged result. Branch left untouched, no push this tick.

**Review gate:** one report-only `/code-review` pass (Standards + Spec axes, parallel).

- **Standards — no hard violations.** Test-only, so no CHANGELOG entry (matches `fe48fa3`, `1bea4a4`). Plain `test_` functions, all signatures annotated, `parametrize` carries explicit `ids=`, no blank line between act and asserts. Two judgement calls, both resolved as fine: the `_unstaged_hunk_id` / `_staged_hunk_id` arrange helpers are plain functions rather than fixtures, matching the de-facto convention already in `dry_run_test.py`, `binary_test.py`, `commit_test.py`; and grouping the cross-cutting banner format into one file avoids repeating the same parametrized shape across four command test files.
- **Spec — all three items of #166 implemented, no scope creep.** The banners are pinned as full-line equality rather than substrings, singular and plural both covered. The deliberately excluded fourth gap (untracked paths can never be staged, overlaps #22) is correctly absent.

No meaningful findings, so `/review-fix` was skipped by design.

**Mutation checks re-run independently** against `git_hunk/_ui.py` (mutated singly, restored between each, tree confirmed clean after):

| Mutation | Result |
| --- | --- |
| `No hunks.` -> `No hunks!` | KILLED |
| `No skills.` -> `No skills found.` | KILLED |
| `print_applied`: `{verb} ` -> `{verb}: ` | KILLED (all 3 params) |
| `print_committed`: drop the plural guard | KILLED (singular test; plural test correctly stayed green) |
| `cmd_show`: `include_untracked=False` -> `True` | KILLED |

**Recommit:** no reshaping needed. Three commits, one per issue item, each independently green: `fdb17d8` 269 passed, `f0632dd` 270 passed, `4c3b1b2` 271 passed.

**Verify:** skipped, test-only PR with no runtime behavior change.

**CI:** all 7 checks green (lint + test across ubuntu/macos/windows x 3.10/3.14). No push happened this tick, so these are authoritative.

No follow-up or idea issues filed: the one loose end in the neighborhood (pinning that an untracked path `list` displays cannot be staged) is already deferred to #22 by the parent issue's own scope note.

PR #168, opened 2026-07-25
test: pin --dry-run on whole-file hunks for discard and unstage

wkentaro:
`_apply_selection` guards the whole-file (binary / mode-only / type-change) branch with a
single shared `if whole_file and not dry_run` (`git_hunk/_cli.py:268`), but only the
`stage` direction of that guard was ever asserted. A regression that dropped it for
`discard` would irreversibly overwrite binary working-tree content during a `--dry-run`
call, and the suite would stay green.

Extracts the binary arrange block into `_setup_binary_modified` (now shared by three
tests) and pins the `discard` and `unstage` directions.

## Test plan
- [x] tests added: `tests/e2e/dry_run_test.py`
- [x] mutation check: replacing the guard with `if whole_file:` fails all three whole-file dry-run tests, including both new ones
- [x] `make lint` and `uv run pytest` (266 passed)

### DIFF
--- a/tests/e2e/dry_run_test.py
+++ b/tests/e2e/dry_run_test.py
@@ -57,19 +57,47 @@ def test_dry_run_line_selection_changes_nothing(cli: GitHunkCLI) -> None:
     assert cli.repo.git("diff", "--cached").strip() == ""
 
 
-def test_stage_dry_run_leaves_binary_unstaged(cli: GitHunkCLI) -> None:
+def _setup_binary_modified(cli: GitHunkCLI) -> tuple[Path, str]:
     path = Path(cli.repo.path) / "a.bin"
     path.write_bytes(b"\x00\x01bin\xff")
     cli.repo.git("add", ".")
     cli.repo.git("commit", "-m", "init")
     path.write_bytes(b"\x00\x02BIN\xfe")
+    hunks = cli.run_list_json("list", "--unstaged", "--json")
+    assert len(hunks) == 1
+    return path, hunks[0]["id"]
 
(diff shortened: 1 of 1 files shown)


wkentaro:

Test-only PR pinning the `discard`/`unstage` directions of the shared `if whole_file and not dry_run` guard in `_apply_selection`.

**Rebase:** not needed. `main` has gained 0 commits since the merge-base and merge state is `CLEAN`, so the existing green still reflects the merged result. Branch left untouched.

**Review:** `/code-review` (medium, report-only) across both axes returned no meaningful findings.
- *Standards*: clean. The helper extraction is justified (3 call sites, not 1), annotations match the sibling `_setup_modified`, and the blank-line placement follows this file's own pre-existing convention. No changelog entry is correct for a test-only change, matching precedent.
- *Spec*: matches the PR description exactly, no scope creep (only `tests/e2e/dry_run_test.py` touched).

**Mutation check re-run independently**, not taken on trust: replacing the guard with `if whole_file:` fails all three whole-file dry-run tests, including both new ones. Reverting restores green at 266 passed. The new tests are load-bearing pins, not vacuous.

**CI:** all 7 checks green (lint + test on ubuntu/macos/windows x 3.10/3.14). No push was made this tick.

One non-blocking note, deliberately not filed as an issue: `test_unstage_dry_run_leaves_binary_staged` asserts only that `git diff --cached` is non-empty rather than naming `a.bin`. It is not vacuous (the mutation run confirms it discriminates, and `a.bin` is the only file in the fixture) and it is directionally symmetric with the existing `== ""` stage assertion, so sharpening it would be churn.

No follow-up or idea issues filed.

PR #169, opened 2026-07-25
test: pin the multi-hunk line-selection guard for unstage, discard and commit

wkentaro:
`_apply_line_filter`'s `"line selection requires exactly one hunk"` guard
(`git_hunk/_cli.py:221`) is shared by all four selection verbs — `stage`,
`unstage`, `discard` and `commit` all reach it through `_apply_selection` — but
only the `stage` direction was ever asserted (`error_test.py`,
`path_shorthand_test.py`). The other three were executed by the suite and
checked by nothing.

Without the guard, a multi-hunk selection carrying `-l` does not error: it
silently drops every hunk but the first and applies the line filter to that one
alone. For `discard` that is unrequested destruction of uncommitted work, and
for `commit` it is a commit the user did not ask for.

Each test pins both halves of the contract — the error is raised, *and* the
relevant state is untouched (working tree for `discard`, index for `unstage`,
`HEAD` for `commit`).

## Test plan
- [x] tests added: `tests/e2e/line_selection_guard_test.py` (3 tests)
- [x] `make test` — 267 passed
- [x] `make lint`
- [x] mutation-tested: with the guard patched out, all three fail; they also
      fail on the state assertions alone with the error assertions stripped, so
      neither half is vacuous

### DIFF
--- a/tests/e2e/line_selection_guard_test.py
+++ b/tests/e2e/line_selection_guard_test.py
@@ -1,0 +1,63 @@
+from pathlib import Path
+
+import pytest
+
+from .conftest import GitHunkCLI
+
+# _apply_line_filter's "exactly one hunk" guard is shared by stage, unstage,
+# discard and commit; only the stage direction is pinned (error_test.py).
+# Body line numbers of the first hunk: 1=" line1" 2="-line2" 3="+CHANGED2",
+# so -l 2,3 would apply to that hunk alone if the guard ever stopped firing.
+
+
+@pytest.fixture
+def two_hunks(cli: GitHunkCLI) -> GitHunkCLI:
+    lines = [f"line{i}" for i in range(1, 21)]
+    cli.repo.write_file("f.py", "\n".join(lines) + "\n")
+    cli.repo.git("add", ".")
+    cli.repo.git("commit", "-m", "init")
+
+    lines[1] = "CHANGED2"
+    lines[17] = "CHANGED18"
+    cli.repo.write_file("f.py", "\n".join(lines) + "\n")
+    return cli
+
+
+def _both_ids(cli: GitHunkCLI, *flags: str) -> list[str]:
+    hunks = cli.run_list_json("list", *flags, "--json")
+    assert len(hunks) == 2
+    return [h["id"] for h in hunks]
+
+
+def test_discard_rejects_line_selection_across_two_hunks(
+    two_hunks: GitHunkCLI,
+) -> None:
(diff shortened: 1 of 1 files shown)


wkentaro:

**Rebase:** not needed. `origin/main` has gained **0 commits** since the merge-base, so the branch sits exactly on top of current `main` and merge state is `CLEAN`. The existing green CI already reflects the merged result; no push was made this tick.

**Review:** `/code-review` (Standards + Spec axes, parallel) returned **no meaningful findings**, so the full `/review-fix` loop was skipped.

- *Standards*: no hard violations. Matches sibling e2e style — no blank line between act and asserts, fully annotated signatures including `-> None`, plain `test_` functions with a fixture, and `_both_ids` correctly renamed from the sibling `_only_id` pattern rather than copy-pasted. The `two_hunks` fixture inlines the arrange block instead of introducing a competing shared helper, which is the right call on a surgical PR. Untouched `CHANGELOG.md` matches the precedent set by prior test-only commits (`fe48fa3`, `1bea4a4`) — test additions are not user-facing. The one baseline smell noted (mild duplicated shape across the three test bodies) mirrors the existing `pattern_selection_test.py` precedent and is not worth extracting at three call sites.
- *Spec*: every claim in the description verified independently. The guard at `_cli.py:221-222` is genuinely shared by all four verbs through `_apply_selection` (line 259); the asserted string matches exactly; and the "only `stage` was covered" claim holds — `"exactly one hunk"` previously appeared only in `error_test.py:150` and `path_shorthand_test.py:100`, both via `stage`. Each state snapshot is taken at the correct moment (worktree before the call, index after staging, `HEAD` before the commit attempt).

**Mutation test reproduced independently:** with the guard removed, all three new tests fail — `discard` returned 0 printing `discarded 5755c57 f.py`, `unstage` returned 0, `commit` returned 0 printing `committed 1 hunk`. That is precisely the silent-partial-operation failure mode the description names, so the pins are load-bearing, not vacuous. Guard restored and all three pass again.

**`/recommit`:** no reshaping needed — one focused commit whose message accurately describes the change. Branch left untouched.

**`/verify`:** skipped; test-only PR with no runtime behaviour change.

**CI:** all 7 checks green (lint + test across ubuntu/macos/windows × 3.10/3.14). Since no push was made this tick, these checks are authoritative and the verdict is issued now rather than waiting.

follow-ups: #170 (the two sibling guards from this same seam — `_build_selection`'s mutual-exclusion checks, asserted only via `stage`, and `_select_hunks`' empty-arg guard, asserted only via `discard`). Does not block this merge.

PR #96, opened 2026-07-08
test: cover CliGroup KeyboardInterrupt exit code 130

wkentaro:
`CliGroup.invoke`'s `except KeyboardInterrupt: ctx.exit(130)` handler (the standard 128+SIGINT convention scripts rely on for Ctrl-C) had no test coverage. This adds a unit test that drives a throwaway command on a fresh `CliGroup` so the interrupt path is exercised directly, without coupling to any real command's internals.

## Test plan
- [x] `uv run pytest tests/unit/_cli/keyboard_interrupt_test.py` passes; mutation-checked (removing the handler makes it fail with exit code 1)
- [x] `uv run ruff check` / `uv run ty check` clean

PR #171, opened 2026-07-25
test: pin shared CLI guards for every verb that reaches them

### LINKED ISSUE #170: Test: finish pinning the shared-guard seam (mutual-exclusion and empty-arg guards asserted for one verb only)

Follow-up spun off from #169, which pinned `_apply_line_filter`'s `"line selection requires exactly one hunk"` guard for `unstage`, `discard` and `commit` (previously only `stage` was asserted).

Two sibling guards in `_cli.py` have the same shape — one guard shared by several verbs, but only **one** verb's direction asserted — and are still unpinned on `main`:

**1. `_build_selection`'s mutual-exclusion checks (`_cli.py:197-207`)**

- `"choose one of -l, --include-matching, or --exclude-matching"`
- `"--regex requires --include-matching or --exclude-matching"`

Both are reached by `stage`, `unstage`, `discard` and `commit`, but every assertion goes through `stage` only (`tests/e2e/pattern_selection_test.py:138,151,158`).

**2. `_select_hunks`' empty-arg guard (`_cli.py:141-143`)**

`"hunk id or file path must not be empty or whitespace"` is reached by `stage`, `unstage`, `discard` and `commit`, but asserted only via `discard` (`tests/e2e/error_test.py:60-70`).

Note: `_find_hunks_by_ids` has its **own separate** blank-id check (`_cli.py:114`, `"hunk id must not be empty..."`) used by `show`, already covered by `tests/unit/_cli/find_hunks_by_ids_test.py:62`. Do not conflate the two — they are different strings and different call paths.

## Suggested approach

Mirror #169: pin both halves of the contract per verb — the error is raised, **and** the relevant state is untouched (index for `unstage`, working tree for `discard`, `HEAD` for `commit`). Mutation-test before opening the PR (patch the guard out, confirm the new tests fail, revert) so the pins are provably load-bearing rather than vacuous.

Existing style precedent to match: no blank line between act and asserts; `pytest.mark.parametrize` carries explicit `ids=`; avoid adding a fresh copy of the `write_file + git add + git commit -m init` arrange block (inline a local fixture instead of refactoring the ~50 pre-existing copies).

wkentaro:
Two shared guards in `_cli.py` are reached by several verbs but were asserted
through only one verb each, so a regression on any other verb's path would have
gone unnoticed:

- `_build_selection`'s mutual-exclusion checks (`choose one of -l,
  --include-matching, or --exclude-matching`, and `--regex requires ...`):
  shared by `stage`, `unstage` and `discard`, asserted only via `stage`.
- `_select_hunks`' empty-argument guard (`hunk id or file path must not be
  empty or whitespace`): shared by `stage`, `unstage`, `discard` and `commit`,
  asserted only via `discard`.

Each new test pins both halves of the contract: the error is raised, and the
verb's own state is left untouched (index for `stage`/`unstage`, working tree
for `discard`, `HEAD` for `commit`).

One correction to the issue text: it lists `commit` as reaching the
mutual-exclusion guard, but `cmd_commit` builds its `_Selection` directly and
exposes no `-l`/`--include-matching`/`--exclude-matching`/`--regex` flags, so it
never calls `_build_selection`. That test is not writable, and its absence here
is deliberate.

No `CHANGELOG.md` entry: test-only, no user-facing change (matching #96).

Closes #170

## Test plan

- [x] Mutation-tested both guards: patching out `sum(mechanisms) > 1` fails all
  6 new `unstage`/`discard` cases, and removing the `not arg.strip()` check
  fails all 3 new `stage`/`unstage`/`commit` cases. Both reverted, suite green.
- [x] `uv run pytest -q`: 273 passed
- [x] `make lint`
- [x] Each commit independently lints, type-checks and passes the suite in a
  detached checkout

### DIFF
--- a/tests/e2e/error_test.py
+++ b/tests/e2e/error_test.py
@@ -135,6 +135,40 @@ def test_empty_hunk_id_rejected(cli: GitHunkCLI) -> None:
     assert cli.repo.git("diff").strip() != ""
 
 
+@pytest.fixture
+def unstaged_change(cli: GitHunkCLI) -> GitHunkCLI:
+    cli.repo.write_file("f.py", "old\n")
+    cli.repo.git("add", ".")
+    cli.repo.git("commit", "-m", "init")
+    cli.repo.write_file("f.py", "new\n")
+    return cli
+
+
+# `stage`, `unstage`, `discard` and `commit` all resolve targets through the
+# same empty-argument guard, so each must refuse to touch its own state.
+def test_empty_hunk_id_rejected_on_stage(unstaged_change: GitHunkCLI) -> None:
+    r = unstaged_change.run("stage", "")
+    assert r.returncode != 0
+    assert "must not be empty" in r.stderr
+    assert unstaged_change.repo.git("diff", "--cached").strip() == ""
+
+
+def test_empty_hunk_id_rejected_on_unstage(unstaged_change: GitHunkCLI) -> None:
+    unstaged_change.repo.git("add", ".")
+    r = unstaged_change.run("unstage", "")
+    assert r.returncode != 0
+    assert "must not be empty" in r.stderr
+    assert unstaged_change.repo.git("diff", "--cached").strip() != ""
+
+
+def test_empty_hunk_id_rejected_on_commit(unstaged_change: GitHunkCLI) -> None:
+    head = unstaged_change.repo.git("rev-parse", "HEAD")
+    r = unstaged_change.run("commit", "-m", "msg", "")
+    assert r.returncode != 0
+    assert "must not be empty" in r.stderr
+    assert unstaged_change.repo.git("rev-parse", "HEAD") == head
+
+
 def test_empty_hunk_id_rejected_on_show(cli: GitHunkCLI) -> None:
     cli.repo.write_file("f.py", "old\n")
     cli.repo.git("add", ".")
--- a/tests/e2e/pattern_selection_test.py
+++ b/tests/e2e/pattern_selection_test.py
@@ -168,3 +168,40 @@ def test_empty_include_matching_is_rejected(added_lines: GitHunkCLI) -> None:
     assert r.returncode != 0
     assert "empty match pattern" in r.stderr
     assert added_lines.repo.git("show", ":f.txt") == "keep\n"
+
+
+# `stage`, `unstage` and `discard` share one selection-flag validator, so the
+# rejections above must hold for the reverse verbs too.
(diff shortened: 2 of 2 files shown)


wkentaro:

Reviewed on two axes (standards + spec) against `origin/main` (3c087be). CI is green across all 7 checks (lint, plus test on ubuntu/macos/windows x 3.10/3.14); merge state `CLEAN`.

**No rebase.** `main` has gained 0 commits since the merge-base, so the branch already sits on current `main` and the existing green reflects the merged result.

**Spec (issue #170): faithful, no scope creep.** The review independently verified this PR's correction to the issue text, and the correction is right: `cmd_commit` declares only `-m`, `-l` and `-h`, and builds its `_Selection` directly (`_cli.py:643`) with `include_matching=(), exclude_matching=(), regex=False` hardcoded. It never calls `_build_selection`, so `commit` cannot reach the mutual-exclusion guard and the omitted test is genuinely unwritable. Coverage is complete for what is reachable: the mutual-exclusion guard now pinned for `unstage` and `discard`, the empty-arg guard for `stage`, `unstage` and `commit` (`discard` was pre-existing). Both halves asserted in every new test (error raised **and** state untouched: index for `stage`/`unstage`, `HEAD` for `commit`, working tree for `discard`). No conflation with `_find_hunks_by_ids`' separate blank-id check.

**Mutation testing reproduced independently.** Patching out `sum(mechanisms) > 1` fails the new `unstage`/`discard` cases; removing the `not arg.strip()` check fails the new `stage`/`unstage`/`commit` cases. The pins are load-bearing, not vacuous.

**Standards: no hard violations.** `-> None` on every test, no test classes, no blank line between act and asserts, explicit `ids=` on `parametrize`, and the `unstaged_change` fixture dedups the arrange block rather than adding a fresh copy. Ruff and ty pass.

Two cosmetic nits, deliberately **not** fixed — a force-push for these would re-trigger CI and cost another cycle for no behavioural gain:
- `"flags, message"` comma-string argnames, where `tests/unit/_cli/normalize_path_arg_test.py:9` uses the tuple form `("arg", "expected")`.
- Hoisting the `parametrize` mark to a module-level `_conflicting_flags` shared by two tests has no precedent in the suite, though it is what the "shared across multiple functions" rule sanctions.

`/recommit` was a no-op: two commits, one guard each, one file each, already independently valid. `/verify` skipped as test-only.

follow-ups: #172 (filed against #105's explicitly deferred help-vs-declared-params drift guard; does not block this PR)

Merging remains yours.

PR #173, opened 2026-07-25
test: pin list's plaintext untracked section as bare paths

wkentaro:
`print_hunk_list` passes `show_hunks=False` for exactly one of its three sections
(untracked), which is why an untracked entry renders as a bare path instead of a
hunk line. That shape was unpinned: every untracked test reads `list --json`,
which bypasses this rendering path, and the one plaintext test that does reach it
(`markup_test.py::test_list_renders_bracketed_untracked_path_verbatim`) asserts
only that the path appears verbatim, not the section header, the bare-path-only
shape, or the absence of a hunk line.

Mutation-checked against the two ways the guard can break; both kept the full
suite green before this test:

1. `show_hunks=False` -> `True` for the untracked section: every untracked file
   gains a hunk line carrying an empty id and a bogus `Mode None -> 100644`
   label, since untracked entries have `id=""` and `a_mode=None`.
2. Renaming the `"untracked:"` section header: that string is unpinned, while
   `"staged:"` and `"unstaged:"` are already pinned by the sibling test directly
   above.

## Test plan

- [x] `uv run pytest -q`: 265 passed
- [x] both mutations above turn the new test red, then reverted
- [x] `make lint`

### DIFF
--- a/tests/e2e/list_test.py
+++ b/tests/e2e/list_test.py
@@ -256,6 +256,23 @@ def test_list_plaintext_blank_lines_separate_sections_and_file_groups(
     assert "b.py" in blocks[2]
 
 
+def test_list_plaintext_untracked_section_lists_bare_paths(cli: GitHunkCLI) -> None:
+    # An untracked entry carries no addressable id, so its section is inventory
+    # only: one bare path per file, never a hunk line (which would render an
+    # empty id and a bogus "Mode None -> 100644" label).
+    cli.repo.write_file("f.py", "init\n")
+    cli.repo.git("add", ".")
+    cli.repo.git("commit", "-m", "init")
+
+    cli.repo.write_file("f.py", "changed\n")
+    cli.repo.write_file("untracked.py", "new\n")
+
+    blocks = cli.run_ok("list").strip("\n").split("\n\n")
+    assert len(blocks) == 2
+    assert blocks[0].startswith("unstaged:")
+    assert blocks[1].splitlines() == ["untracked:", "untracked.py"]
+
+
 def test_list_context_before_field_in_json_matches_display(cli: GitHunkCLI) -> None:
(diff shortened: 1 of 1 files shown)


wkentaro:

Test-only change adding one e2e test. No rebase needed (`origin/main` gained 0
commits since the merge-base, merge state `CLEAN`), so the existing green checks
already reflect the merged result. No push this tick; all 7 checks pass (lint,
plus test on ubuntu/macos/windows x 3.10/3.14).

**Standards axis:** clean. Plain `test_` function, annotated `-> None`, no blank
line between act and asserts, `assert len(blocks) == 2` before indexing, and the
comment explains *why* rather than restating the code. The inlined
`write_file` + `add` + `commit` arrange block matches the ~50 existing e2e tests;
refactoring it here would be non-surgical.

**Spec axis:** the claims were re-verified independently rather than taken at
face value. `show_hunks=False` for the untracked section only, the bare-path
branch at `_ui.py:100-101`, and the `Mode None -> 100644` label that a flipped
flag would emit all check out. Both mutations were reproduced from scratch and
each turns the new test red; `uv run pytest -q` gives 265 passed and `make lint`
is green.

One correction came out of that pass: the PR body claimed every untracked test
reads `list --json`, "which bypasses this rendering path entirely". That
overstated the gap, since
`markup_test.py::test_list_renders_bracketed_untracked_path_verbatim` already
reads plaintext `list` for an untracked file. It asserts only that the path
appears verbatim, not the section header, the bare-path-only shape, or the
absence of a hunk line, so the test still adds real coverage. The body has been
edited to say that accurately. No code change was needed.

follow-ups: #174 (the two neighbouring `_ui.py` render paths this PR leaves
unpinned: the untracked N>1 spacing asymmetry, and `print_hunk_diffs`'
inter-hunk divider). Does not block this merge.

PR #175, opened 2026-07-25
test: pin _ui.py's untracked spacing and show's inter-hunk rule

### LINKED ISSUE #174: Test: pin the remaining plaintext render paths in _ui.py (untracked N>1 spacing, show's inter-hunk divider)

Follow-up from #173, which pinned the untracked section's bare-path shape for a
single untracked file. Two neighbouring pieces of `_ui.py`'s plaintext render
path are still executed by the suite but never asserted, so both can be broken
without turning a test red.

### 1. `_print_status_section`: the N>1 untracked shape

`git_hunk/_ui.py:95-101`. The `show_hunks=True` branch separates file groups with
a blank line (`if i < len(by_file) - 1: out.print()`); the `show_hunks=False`
branch (untracked) has no such separator, so multiple untracked files render as
consecutive bare paths with no blank line between them. #173 only exercises N=1,
so that asymmetry is unpinned: adding a blank-line separator to the untracked
branch, or removing it from the tracked branch, keeps the suite green.

### 2. `print_hunk_diffs`: the inter-hunk divider

`git_hunk/_ui.py:167-172`. `if i > 0: out.print(Rule(style="dim"))` draws a
divider between hunks in `show` output. It can be deleted, doubled, or moved
before the first hunk with no test noticing.

### Suggested approach

Two e2e tests, matching the style of `list_test.py::test_list_plaintext_untracked_section_lists_bare_paths`:

- `list` with two untracked files, asserting the exact plaintext block shape.
- `show` with two hunks, asserting exactly one divider and that it falls between
  them, not before the first.

Assert `len(blocks) == N` before indexing, and put no blank line between the act
and the asserts (repo precedent).

Mutation-check each test before opening the PR: patch the guard out, confirm the
new test fails, revert. Note `python` is not on PATH in this repo's shell, use
`uv run python` / `uv run pytest`.

wkentaro:
Closes #174.

Follow-up to #173, which pinned the untracked section for a single file. Two
neighbouring branches of `_ui.py`'s plaintext render path were still executed by
the suite but never asserted, so either could be broken without turning a test
red:

- `_print_status_section`'s `show_hunks=False` branch emits no blank line between
  untracked paths, unlike the tracked branches that separate file groups. #173
  only exercised N=1, so the asymmetry was unpinned.
- `print_hunk_diffs`' inter-hunk rule could be deleted, doubled, moved before the
  first hunk, or reduced to a single rule after the first hunk with no test
  noticing.

The show test renders three hunks rather than two. At N=2 the contract "a rule
between each pair" is indistinguishable from "one rule, ever", so two files would
have reproduced the same N=1 blind spot this PR exists to close.

## Test plan
- [x] tests added: `tests/e2e/list_test.py`, `tests/e2e/show_test.py`
- [x] mutation-checked each guard against the full suite, each mutation failing only the new test:
  - untracked branch: adding a blank-line separator fails the list test
  - inter-hunk rule: deleting it, doubling it, moving it before the first hunk, or firing it only once (`if i == 1`) each fail the show test
- [x] `make test` (326 passed), `make lint`

### DIFF
--- a/tests/e2e/list_test.py
+++ b/tests/e2e/list_test.py
@@ -272,3 +272,20 @@ def test_list_context_before_field_in_json_matches_display(cli: GitHunkCLI) -> N
     assert hunks[0]["context_before"] == {"text": "def foo():"}
     # Parity: the field equals the function context shown in the human display.
     assert "def foo():" in cli.run_ok("list")
+
+
+def test_list_plaintext_untracked_paths_are_not_separated_by_blank_lines(
+    cli: GitHunkCLI,
+) -> None:
+    # The tracked sections put a blank line between file groups; the untracked
+    # section is a flat inventory instead, so its paths stay consecutive.
+    cli.repo.write_file("f.py", "init\n")
+    cli.repo.git("add", ".")
+    cli.repo.git("commit", "-m", "init")
+
+    cli.repo.write_file("a.py", "new\n")
+    cli.repo.write_file("b.py", "new\n")
+
+    blocks = cli.run_ok("list").strip("\n").split("\n\n")
+    assert len(blocks) == 1
+    assert blocks[0].splitlines() == ["untracked:", "a.py", "b.py"]
--- a/tests/e2e/show_test.py
+++ b/tests/e2e/show_test.py
@@ -122,3 +122,24 @@ def test_show_renders_no_newline_marker_unnumbered(cli: GitHunkCLI) -> None:
     assert marker_line.strip() == NO_NEWLINE_MARKER
 
     assert "+cX" in r.stdout
+
+
+def test_show_puts_exactly_one_rule_between_consecutive_hunks(
+    cli: GitHunkCLI,
+) -> None:
+    # The rule divides hunks, so it appears between them and never leads the
+    # output: N hunks render N-1 rules. Three files rather than two, so that a
+    # rule firing once and never again stays distinguishable from the contract.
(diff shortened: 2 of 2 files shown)


wkentaro:

- **Rebase:** not needed. The branch sits exactly on the current `origin/main` tip (0 commits gained since the merge-base, no file overlap), so the existing green already reflects the merged result. Merge state `CLEAN`.
- **Review:** report-only `/code-review` over the diff came back clean on both axes. Standards: 0 findings; both tests match in-repo precedent (the `split("\n\n")` block idiom, `assert len(blocks) == N` before indexing, no blank line between act and asserts, complete annotations). Spec: 0 findings against #174, no scope creep. No escalation to `/review-fix`.
- **Load-bearing check:** the mutation results in the PR description were independently reproduced. Adding a blank-line separator to `_print_status_section`'s `show_hunks=False` branch fails the list test (`len(blocks)` 1 -> 2); dropping the `if i > 0` guard so a rule leads the output fails the show test (`len(rules)` 1 -> 2, one at index 0). Both guards revert cleanly. These are real pins, not vacuous coverage.
- **Commits:** one conventional commit for one cohesive change. No reshaping needed.
- **CI:** all 7 checks green (lint, plus test on ubuntu/macos/windows x 3.10/3.14).
- **Smoke:** skipped, test-only PR with no runtime behaviour change.

No follow-up or idea issues filed.

PR #176, opened 2026-07-26
test: pin submodule gitlink pointer changes as patchable text hunks

wkentaro:
A moved submodule pointer reaches git-hunk as a one-line `Subproject commit` diff carrying mode `160000` on both sides. Nothing in the suite exercised a gitlink, so all of its handling was unasserted — and it is handled entirely by fallthrough: `git_hunk/_hunk.py` has no `160000` branch at all, so a gitlink lands in the ordinary text-hunk path only because `a_mode == b_mode` keeps it out of the whole-file (chmod) branch.

## Summary
- `list --json` reports `a_mode`/`b_mode` `"160000"`, `binary: false`, and a real `@@ -1 +1 @@` header, i.e. an ordinary text hunk rather than a whole-file hunk like a binary or mode-only change
- plaintext `list` renders it with no `Mode` / `Type change` / `Binary file` label, since those are reserved for hunks with no `@@` range
- `show` renders the `Subproject commit` body
- `stage` then `unstage` round-trips it cleanly

The fixture builds the gitlink with a nested `git init sub` plus one commit rather than `git submodule add`, so it needs no `.gitmodules` and no `protocol.file.allow=always` override that a local-path submodule would require on CI. git records both shapes as a mode `160000` gitlink and emits an identical diff, and git-hunk never reads `.gitmodules`.

## Test plan
- [x] tests added: `tests/e2e/submodule_test.py` (4 tests)
- [x] full suite green: 268 passed
- [x] mutation-tested to confirm the tests are load-bearing, not vacuous: skipping `b_mode == "160000"` blocks in `parse_diff` fails exactly the 3 JSON/round-trip tests and nothing else; giving gitlinks a special label in `_ui._header_text` fails exactly the new plaintext test

## Follow-up (not addressed here)
Found while writing these tests, and left alone deliberately since it needs a product call: `git-hunk stage <gitlink-id> -l 1` leaks a raw git error rather than a clean `CliError`.

```console
$ git-hunk stage 2c987d7 -l 1
error: git apply --whitespace=nowarn --cached failed: error: corrupt patch for
submodule sub
```

Because a gitlink hunk has a non-empty `diff`, `_is_whole_file_hunk` is false, so `_apply_line_filter` lets `-l` through instead of raising the `"line selection is not supported for binary, mode, or type changes"` error that binary/mode/type hunks get. Whether `160000` should join that rejection list is a maintainer decision, so this PR neither fixes it nor pins the current behavior as correct.

### DIFF
--- a/tests/e2e/submodule_test.py
+++ b/tests/e2e/submodule_test.py
@@ -1,0 +1,79 @@
+import pytest
+
+from .conftest import GitHunkCLI
+
+
+def _bump_submodule(cli: GitHunkCLI, content: str) -> None:
+    cli.repo.write_file("sub/f", content)
+    cli.repo.git("-C", "sub", "add", "f")
+    cli.repo.git("-C", "sub", "commit", "-m", content)
+
+
+@pytest.fixture
+def bumped_submodule(cli: GitHunkCLI) -> GitHunkCLI:
+    # An embedded repository, not a .gitmodules-registered submodule: git records
+    # both as a mode 160000 gitlink and emits the same "Subproject commit" diff,
+    # and git-hunk never reads .gitmodules, so this skips `git submodule add` and
+    # the protocol.file.allow override a local-path submodule would need on CI.
+    cli.repo.git("init", "sub")
+    cli.repo.git("-C", "sub", "config", "user.email", "test@test.com")
+    cli.repo.git("-C", "sub", "config", "user.name", "Test")
+    _bump_submodule(cli, "one")
+    cli.repo.git("add", "sub")
+    cli.repo.git("commit", "-m", "init")
+    _bump_submodule(cli, "two")
+    return cli
+
+
+def test_list_reports_a_gitlink_bump_as_a_text_hunk(
+    bumped_submodule: GitHunkCLI,
+) -> None:
+    hunks = bumped_submodule.run_list_json("list", "--unstaged", "--json")
+    assert len(hunks) == 1
+    hunk = hunks[0]
+    assert hunk["file"]["text"] == "sub"
+    assert hunk["change_kind"] == "M"
+    assert hunk["a_mode"] == "160000"
+    assert hunk["b_mode"] == "160000"
+    # A moved submodule pointer is a one-line "Subproject commit" diff, so it is
+    # a patchable text hunk rather than a whole-file one despite the odd mode.
+    assert hunk["binary"] is False
+    assert hunk["header"] == "@@ -1 +1 @@"
+    assert hunk["additions"] == 1
+    assert hunk["deletions"] == 1
+
+
+def test_list_plaintext_gives_a_gitlink_no_whole_file_label(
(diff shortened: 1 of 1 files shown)


wkentaro:

## Rebase

Skipped, and no push was made this tick. The branch's merge-base is the current `origin/main` tip (`git rev-list --count $base..origin/main` = **0**), merge state is `CLEAN`, and all 7 checks are green, so the existing CI result already reflects the merged tree. Nothing to re-trigger.

## Review

A report-only `/code-review` gate ran over the diff on both axes. It did not escalate to `/review-fix`.

**Spec — 0 findings.** Every claim in the PR body was verified rather than trusted:

- `uv run pytest tests/e2e/submodule_test.py -q` → 4 passed; full suite → **268 passed**, matching the stated number.
- The fixture claim was reproduced from scratch in a scratch repo: `git init sub` + commit + `git add sub` yields byte-for-byte the described diff (`160000` both sides, one `@@ -1 +1 @@`, `Subproject commit` body) and creates **no `.gitmodules`**. The "no `protocol.file.allow` override needed" claim holds.
- `git_hunk/_hunk.py` confirmed to contain **no `160000` special-casing at all**, so the fallthrough rationale in the description is accurate, not a post-hoc story.
- `git_hunk/_ui.py`: `_header_text` returns `hunk.header` whenever it is non-`None`, so the `Mode` / `Type change` / `Binary file` label path is *structurally* unreachable for a gitlink. The negative assertions in `test_list_plaintext_gives_a_gitlink_no_whole_file_label` are anchored by a positive `"@@ -1 +1 @@" in out` check, so they cannot pass vacuously against empty output.

**Standards — 1 finding, judgement call, deliberately not actioned.** The reviewer flagged the repeated three-line "look up the single hunk's id" block and suggested extracting an `_only_id`-style helper, citing sibling files as precedent. That precedent is real but points the other way here: `_only_id` is already copy-pasted verbatim at `tests/e2e/line_selection_test.py:21` and `tests/e2e/pattern_selection_test.py:20`, so adding a third copy would deepen the existing duplication rather than resolve it. Consolidating those into a shared fixture is a suite-wide refactor, out of scope for a surgical test-only PR. Everything else conformed: annotations throughout, no test classes, aspect-based filename, length asserted before every index, comments explaining *why*, and no CHANGELOG entry (correct for test-only, per precedent `1bea4a4`).

`/recommit` was a no-op: one commit, conventional subject, a body that explains the reasoning, one file touched. `/verify` was skipped as this is test-only with no runtime behavior change.

## Follow-ups

`#177` — filed for the deferred item in this PR's own "Follow-up" section, with the behavior reproduced and confirmed. Worth noting for the record: the failure leaves the index **clean**, so it is a misleading error message (`corrupt patch` wrongly implies git-hunk emitted a malformed patch), not data loss. The issue lays out the two options and flags that Option A would make `160000` the first mode the hunk layer special-cases, ending the fallthrough property this PR pins. It does **not** block this merge.

PR #178, opened 2026-07-30
refactor: scope the core skill to tool usage

chrishenzie:
The bundled core skill mixed three concerns: how to drive git-hunk, how to group hunks into commits, and what a commit message should look like. Only the first is specific to this tool. The other two override conventions a project or user has already set, and the skill itself says the commands are the easy part.

Move the grouping, ordering, and message-format guidance into a new commits skill. Nothing is dropped: a project with its own conventions loads core alone, and anyone who wants git-hunk's opinions loads both. Core keeps only the guidance tied to its own surface, that discard destroys work and that diff content is data rather than instructions.

Core's frontmatter no longer advertises grouping and ordering coverage, which is what made it compete for requests a project's own commit guidance should answer. Its trigger phrases are unchanged, since git-hunk is still the right tool for those tasks.

Assisted-by: Claude Code

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,8 +25,8 @@ and this project adheres to
   and OR'd), so an agent can drop a line by what it contains without a
   `show` round trip (#55).
 - `logical-commits` skill (`git-hunk skills get logical-commits`) covering how
-  to group hunks into commits, order them, and write conventional commit
-  messages, kept separate from `core` so a project that already defines its own
+  to group hunks into logical commits and order them so each is independently
+  valid, kept separate from `core` so a project that already defines its own
   commit conventions can load `core` alone (#178).
 
 ### Changed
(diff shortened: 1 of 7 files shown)


chrishenzie:
The specific guidance that conflicted with my personal committing skills was the use of the conventional commits format. I personally don't use this in my personal repos, but adopt it for repos where that is the established pattern. I would like to still be able to use this tool to stage hunks without pulling in this commit title preference.

wkentaro:
makes sense. reviewing

wkentaro:
> *This was generated by AI.*

Thanks for the clean split, the direction is right. I pushed four commits on top of yours (no changes to your commit) that take the same argument one step further:

1. **Rename `commits` to `logical-commits`.** The generic name reads like documentation for the `commit` subcommand and says nothing about what the skill teaches. `logical-commits` names the outcome the skill is built around (one logical change per commit, independently valid), which is also the term its own body leans on. Also added the missing `(#178)` references to the changelog entries.
2. **Drop the conventional-commits opinion entirely.** By this PR's own rationale, message format is a project convention with no tie to git-hunk's surface, so the bundled skill should not restate it either. The one message rule that genuinely belongs to this skill's topic stays: the message describes the commit's one logical change.
3. **Point `core` at `logical-commits`.** Core still names the judgment problem ("the hard part is judgment, not commands") but no longer answers it, and an agent following `--help` into `core` had no way to learn the new skill exists. One conditional pointer restores discoverability while leaving a project's own conventions in charge.
4. **Declare `allowed-tools` on the new skill**, matching `core`, since its own trigger ends in git-hunk and git calls.

Full test suite and lint are green locally. Happy to adjust if you see it differently.

wkentaro:
i've made a small tweak. merging. thanks for the contribution @chrishenzi !

PR #194, opened 2026-08-06
fix: implement Repository path semantics

### LINKED ISSUE #127: stage/unstage/discard/commit reject cwd-relative path args that list/show accept from a subdirectory
> *This was generated by AI during a kaizen pass.*

## Summary

`stage`/`unstage`/`discard`/`commit <path>` do not resolve a **cwd-relative** path argument the way `list`/`show <path>` (and plain `git`) do. Run from a subdirectory, the same relative path that `list` accepts is rejected by `stage` with "no changed file matches".

## Reproduction (confirmed against `origin/main` @ 3c087be)

```console
$ git init -q repro && cd repro
$ mkdir sub && printf 'one\n' > sub/nested.py && git add -A && git commit -qm init
$ printf 'two\n' > sub/nested.py
$ cd sub

$ git-hunk list nested.py       # works: git resolves the path relative to cwd
unstaged:
sub/nested.py
  5241625  @@ -1 +1 @@  +1 -1

$ git-hunk stage nested.py      # fails on the same argument, same directory
error: no changed file matches 'nested.py'
```

## Root cause

`list`/`show` forward the path straight into `git diff -- <path>` (`_get_hunks` -> `get_diff(files=...)`), so **git** resolves it relative to cwd. `stage`/`unstage`/`discard`/`commit` instead fetch the whole diff and match in-process in `_select_hunks` (`git_hunk/_cli.py:135-160`): `_normalize_path_arg(arg)` is compared against `h.file`, which is always repo-root-relative. From `sub/`, `_normalize_path_arg("nested.py") == "nested.py"` never equals `"sub/nested.py"`, so it falls through to the "no changed file matches" branch. `_normalize_path_arg` (added in #95) only normalizes separators; it does not account for the cwd prefix.

## Why this needs a maintainer decision (not a surgical PR)

This is a real inconsistency, but the fix has more than one reasonable shape and lands in path-handling code that is **actively being reworked**:

1. **What is the intended contract for path args to `stage`/`unstage`/`discard`/`commit`?** cwd-relative pathspecs (git-familiar, matches `list`/`show`), or repo-root-relative (matches the `--json` `file.text` basis)? Both are defensible; the current behavior is a third, unintended thing (repo-root-relative but only when you happen to run from the root).
2. **Which implementation?** (a) resolve the arg to repo-root-relative via the git prefix (`git rev-parse --show-prefix`) inside `_select_hunks`, or (b) delegate to git as a pathspec the way #103 chose for the untracked filter ("let git resolve the file-path filter as a pathspec"). Option (b) is more consistent with the in-flight direction but changes how `stage` fetches hunks (it currently fetches all, because args mix hunk-ids and paths), so the two resolution paths need reconciling.
3. **Interaction with open PR #103** ("report untracked files with repo-root-relative paths"), which is reworking exactly this cwd-vs-root path handling and adds `tests/e2e/subdirectory_test.py`. A fix here should align with the approach that lands there rather than pre-empt it.

## Affected files

- `git_hunk/_cli.py` -- `_select_hunks` (`:135-160`), `_normalize_path_arg` (`:128-134`); shared by `cmd_stage`/`cmd_unstage`/`cmd_discard`/`cmd_commit`.
- Related in-flight: PR #103, PR #95 (merged, origin of `_normalize_path_arg`).

## Suggested test once a direction is chosen

e2e test that `chdir`s into a subdirectory, then asserts `git-hunk stage <cwd-relative-path>` succeeds the same way `git-hunk list <cwd-relative-path>` already does (current e2e tests all run from the repo root, so this case is untested).

### LINKED ISSUE #159: Silent data loss: textual stage/unstage/discard no-ops (reports success) when run from a repo subdirectory
> *This was generated by AI during a kaizen pass.*

## Summary

Run from a repo subdirectory, `git-hunk stage`/`unstage`/`discard` of a **textual** hunk whose file lives outside that subdirectory **silently no-ops**: the CLI prints its normal `staged ...` success line, exits `0`, but the index is never touched. This is a data-integrity bug (fabricated success), and unlike #127 it needs no path argument at all — a bare **hunk-id** invocation reproduces it.

## Reproduction (confirmed against `origin/main` @ 3c087be)

```console
$ git init -q repro && cd repro
$ mkdir dirA dirB
$ printf 'a1\n' > dirA/a.txt && printf 'b1\n' > dirB/b.txt
$ git add -A && git commit -qm init
$ printf 'a1\na2\n' > dirA/a.txt && printf 'b1\nb2\n' > dirB/b.txt

$ git-hunk list            # from root: both hunks visible, dirB hunk id e3e50a0
unstaged:
dirA/a.txt
  3e82b31  @@ -1 +1,2 @@  +1
dirB/b.txt
  e3e50a0  @@ -1 +1,2 @@  +1

$ cd dirA
$ git-hunk stage e3e50a0   # reports success...
  staged e3e50a0  dirB/b.txt  @@ -1 +1,2 @@  +1
$ echo $?
0
$ cd .. && git diff --cached --stat   # ...but nothing was staged
$ git status --short
 M dirA/a.txt
 M dirB/b.txt               # dirB/b.txt still fully unstaged
```

## Root cause

`run_git` (`git_hunk/_git.py:8-20`) shells out to `git` with no working-directory anchoring (no `cwd=`, no `git -C <toplevel>`; `--show-toplevel` appears nowhere in the codebase). Meanwhile `get_diff`/`parse_diff` always produce **repo-root-relative** file paths (git's own behavior), and `Hunk.file` / the rebuilt patch use those verbatim.

`git apply` (used by `apply_patch`, `_git.py:41-55`, to stage/unstage/discard textual hunks) has a documented behavior: *"When running from a subdirectory in a repository, patched paths outside the directory are ignored"* — silently, exit `0`, no warning. So from `dirA/`, the patch for `dirB/b.txt` is dropped, yet `apply_patch` returns normally and the CLI reports success.

Note the split failure mode from the same missing-cwd defect: **whole-file** hunks (binary/mode/type changes, untracked adds) go through `git add --`/`git restore --` instead, which are also cwd-relative but fail *loudly* (`pathspec ... did not match`, surfaced as `RuntimeError`/`CliError`). So the same root cause yields loud failure for whole-file hunks and **silent** success-shaped data loss for textual hunks — the silent one being the dangerous case.

## Relationship to existing work

- **Distinct from #127** ("stage/unstage/discard reject cwd-relative path *args*"): #127 is about `_normalize_path_arg`/`_select_hunks` rejecting a user-supplied *file argument*, and it fails **loudly**. This bug is in the git subprocess's working directory, fires with **no path arg** (hunk-id only), and fabricates **silent** success.
- **Same subdirectory theme** as #127 (parked as `needs-triage`, a product decision on the intended path contract) and in-flight **PR #103** (reworks cwd-vs-root path handling, adds `tests/e2e/subdirectory_test.py`). A fix here should be reconciled with whatever contract those settle on rather than pre-empting it.

## Why this needs a maintainer decision (not a surgical PR)

1. **Blast radius.** The natural fix — resolve the repo toplevel once (`git rev-parse --show-toplevel`) and pass `cwd=` on every `subprocess.run` in `_git.py` — changes the working directory of *every* git call in the tool (diff, apply, add, restore, commit, ls-files, rev-parse). That is the most central behavior to change unattended.
2. **Scope has more than one reasonable answer.** (a) Minimal: anchor `run_git` to the toplevel so hunk-id ops become correct from any subdir, leaving #127's path-arg contract untouched; or (b) resolve subdirectory support holistically alongside #127/#103 so path-arg and hunk-id invocations behave consistently. Shipping (a) alone leaves a half-supported state (id-based works from a subdir, path-arg still rejects per #127), which the maintainer may consider worse than the current uniform non-support.
3. **Amends a parked decision.** Subdirectory handling was deliberately parked (#127); advancing it here is a product call, not a surgical cleanup.

## Candidate fix (for whoever picks this up)

Resolve `git rev-parse --show-toplevel` once and pass it as `cwd=` to every `subprocess.run` in `run_git` (or prefix every invocation with `git -C <toplevel>`). At the repo root this is a no-op (toplevel == cwd), so it cannot regress the from-root case.

## Suggested regression test

The e2e fixture (`tests/e2e/conftest.py`) always `os.chdir`s to the repo root, so no test exercises a nested-subdir invocation. Add an e2e test that creates files in two sibling subdirectories, `os.chdir`s into one, and asserts `git-hunk stage <id-of-the-other-dir-file>` actually stages it (assert on `git diff --cached`). It fails today (silent no-op) and passes once `run_git` is anchored to the toplevel.

## Affected files

- `git_hunk/_git.py` — `run_git` (`:8-20`), `apply_patch` (`:41-55`), `stage_files`/`unstage_files`/`discard_files` (`:58-67`).
- Related: #127 (parked), PR #103 (in flight).

### LINKED ISSUE #187: Implement Repository path semantics
> *This was generated by AI.*

## Agent brief

Implement the Repository path contract for v0.3.0. Resolve the shared path boundary before other file-state work starts.

## Decision and plan

- [Repository path decision](https://github.com/wkentaro/git-hunk/issues/183#issuecomment-5199959919)
- [Exact release checklist](https://github.com/wkentaro/git-hunk/blob/45758d791edbb207dc1682df04aaf45af440ce5d/docs/research/186-v0.3.0-release-candidate-checklist.md#1-implement-repository-path-semantics)

## Scope

Resolve these source issues in one implementation:

- [stage, unstage, discard, and commit reject cwd-relative path arguments](https://github.com/wkentaro/git-hunk/issues/127)
- [Text operations silently no-op from a repository subdirectory](https://github.com/wkentaro/git-hunk/issues/159)

Use one exact, root-relative Repository path coordinate system. Root-anchor every Git query and mutation. Force canonical diff paths. Reject absolute and escaping operands. Do not expand directories, globs, or Git pathspec magic.

## Acceptance

- Add the full end-to-end matrix from the decision artifact.
- Assert exact HEAD, index, worktree bytes, modes, and status. A success banner is not sufficient.
- Update all affected public docs, bundled skills, and the Unreleased changelog with this pull request number.
- The pull request must close both source issues.
- Run focused tests, full tests, and full lint.

wkentaro:
> *This was generated by AI.*

## Summary

Give git-hunk one path coordinate system, so every command behaves the same from
the worktree root and from any directory below it.

A **Repository path** is relative to the worktree root, uses `/`, and has the same
meaning from every invocation directory. It is now the only path basis: `Hunk.file`,
plain output, success output, internal patches, whole-file mutations, and every user
file operand.

- **Root-anchor every git call.** Bootstrap once with `rev-parse --show-toplevel`,
  then run every later query and mutation there. `git apply` silently ignores patched
  paths outside its working directory, which is what made a text `stage` from a
  subdirectory print success, exit `0`, and never touch the index.
- **Exact literal operands.** An operand names one exact changed Repository path.
  Directories, globs, and pathspec magic are not expanded. A leading `./` and internal
  `..` are normalized; absolute and escaping paths are rejected with a tip naming the
  worktree root.
- **Canonical diff paths.** `--no-relative --src-prefix=a/ --dst-prefix=b/` pins the
  output basis, so `diff.relative`, `diff.noprefix`, and `diff.mnemonicPrefix` cannot
  shift the paths or the Hunk IDs derived from them.
- **Repository before operand.** A Repository path has no meaning outside a worktree,
  so an invocation outside one reports `not a git repository` rather than complaining
  about the shape of a path.

The decision is recorded in [ADR 0002](docs/adr/0002-repository-path.md).

## Breaking changes

- A file operand is root-relative everywhere. From `sub/`, `same.txt` selects the
  worktree-root file; `sub/same.txt` selects the one inside `sub/`.
- Minimum supported Git is now **2.28** (`git diff --no-relative`). `--default-prefix`
  would have been the direct spelling but requires 2.41, which excludes Ubuntu 22.04
  LTS (2.34) and Debian 12 (2.39).
- A pathspec-shaped operand such as `':(bogus)x'` no longer reaches git, so it no
  longer produces the clean git-failure error added in #107. Mutation commands report
  `no changed file matches`; `list` returns an empty inventory.
- An all-whitespace operand is now a legal Repository path.

## Notes

`--literal-pathspecs` is passed only by `git add` / `git restore`, the calls that hand
git a path as a pathspec. Setting it globally would export `GIT_LITERAL_PATHSPECS=1` to
every child process, which silently changes how a commit hook's own pathspecs behave.

Mid-mutation atomicity (a failure between the `git apply` leg and the whole-file leg)
stays out of scope here and belongs to the file-level edge-case work.

## Test plan

- [x] `make lint`
- [x] `make test` (445 passed)
- [x] Each of the three commits lints and tests green in a detached checkout
- [x] New `tests/e2e/repository_path/` suite runs from a subdirectory and asserts exact
      `HEAD`, index, worktree bytes, modes, object IDs, and `status --porcelain=v1 -z`
      after each command, under hostile `diff.*` config. A success banner is never
      treated as evidence.
- [x] Covered: root/subdirectory name collisions, names containing glob and pathspec
      punctuation, text and whole-file `stage`/`unstage`/`discard`/`commit` by ID and
      by path, type changes, `--dry-run`, mixed text + whole-file selections, and
      operands rejected outside the Repository path space
- [x] Punctuation-path tests mutation-checked: they fail if `--literal-pathspecs` is
      removed
- [x] Verified a `pre-commit` hook no longer inherits `GIT_LITERAL_PATHSPECS`

Closes #187
Closes #127
Closes #159


wkentaro:

Verdict: recommend merge.

- Standards: no findings remain after the review fixes. Normal Git operations now require the worktree root, and the new prose follows the repository style rules.
- Spec: the change implements the Repository path contract from #187 and covers #127 and #159 without scope creep.
- Verification: `make lint` passed, all 516 tests passed, and each of the three commits passed independently.
- CI: all seven Linux, macOS, Windows, Python 3.10, and Python 3.14 checks passed after the rebase onto current `main`.
- Merge state: clean and mergeable.

wkentaro:

## Independent review verdict: `recommend-merge`

Reviewed the full diff (code, tests, docs) and re-verified the test plan locally.

**Verified:**

- `make lint` and `make test` pass at HEAD (528 passed).
- The middle commit (`d9e5f16`) is independently green (528 passed in a detached worktree).
- The mutation-check claim holds: neutralizing `--literal-pathspecs` in `_git.py` makes `test_stage_binary_repository_path_with_pathspec_punctuation` fail, so the punctuation tests really guard the flag.

**Review notes (none blocking):**

- The design is coherent: one Repository path basis, every git call anchored at `rev-parse --show-toplevel`, canonical diff prefixes pinned against hostile `diff.*` config. `GitCommandError` stays a `RuntimeError` subclass, so existing handlers still catch it, and the bare-repo test asserts git's own stderr survives as the tip.
- The new e2e suite asserts exact post-state (HEAD id, index and worktree bytes, modes, object ids, `status --porcelain -z`) instead of trusting the success banner, which is exactly the right lesson from #159. Root/subdirectory name collisions, pathspec punctuation, mode changes, type changes, dry-run, and mixed selections are all covered.
- `_select_hunks` tests the raw argument (not the normalized path) for hex, so `./d161935` never resolves as an ID. That reads as deliberate and is the safer choice.
- `list` now diffs the whole repository and filters in-process rather than passing a pathspec to git. A minor cost on very large repos, and the correct trade for exact-literal semantics.
- The PR body says 445 passed; the branch now passes 528. Just the suite growing after the description was written.

PR #197, opened 2026-08-06
test: ignore an inherited git environment

wkentaro:
Running the suite from inside `git rebase --exec` commits each fixture's setup
into the repository under test. I hit this verifying a branch commit-by-commit;
it injected ~56 fixture commits before the rebase aborted.

`GitRepo.run` passes `cwd=`, but git exports `GIT_DIR` and `GIT_INDEX_FILE` to
the commands it runs, and those beat `cwd`. Hooks, `filter-branch`, and
`bisect run` export them too.

Two non-obvious bits:

1. The scrub is session-scoped rather than a per-call `env=` on `GitRepo.run`,
   because the leak also reaches the git subprocesses `git_hunk` itself spawns.
   Fixing it at the fixture would have meant threading an env through
   production code for a test concern.

2. It drops all of `GIT_*` rather than a denylist; a denylist misses
   `GIT_OBJECT_DIRECTORY`, `GIT_COMMON_DIR`, `GIT_NAMESPACE`, and the
   `GIT_AUTHOR_*` / `GIT_COMMITTER_*` values `rebase` also exports. The one
   exception is the config overrides a caller may set to shield the suite from
   the machine's gitconfig (`GIT_CONFIG_GLOBAL`, `GIT_CONFIG_SYSTEM`,
   `GIT_CONFIG_NOSYSTEM`); those improve hermeticity, so the scrub keeps them.

`tests/git_env_test.py` points a decoy `GIT_DIR` at a throwaway repo and asserts
its commit count stays zero (`git rev-list --all --count`, which also exits
non-zero if the decoy path is wrong). A bare "no `GIT_*` in `os.environ`"
assertion would have passed in a normal local run and given a false green.

## Test plan

- [x] `git rebase --exec 'make test' main` completes clean at every commit, HEAD unchanged
- [x] full suite under poisoned `GIT_DIR` / `GIT_INDEX_FILE` / `GIT_WORK_TREE`:
      358 passed, repository untouched
- [x] suite passes with `GIT_CONFIG_GLOBAL=/dev/null` and `GIT_CONFIG_NOSYSTEM=1`
      kept through the scrub, decoy `GIT_DIR` still dropped
- [x] regression test fails with the fixture removed
- [x] `make test` (358 passed) and `make lint`

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -34,6 +34,20 @@ class GitRepo:
         return filepath
 
 
+@pytest.fixture(autouse=True, scope="session")
+def _scrubbed_git_env() -> Generator[None]:
+    # git exports GIT_DIR, GIT_INDEX_FILE, and friends to the commands it runs:
+    # `rebase --exec`, hooks, `filter-branch`, `bisect run`. Those beat cwd, so
+    # every git subprocess the suite starts (the fixtures' own, and the ones
+    # git_hunk spawns) would target the outer repository instead of the
+    # temporary one, and the tests would commit into the repository under test.
+    # Nothing here wants an inherited git environment, so drop all of it.
+    with pytest.MonkeyPatch.context() as monkeypatch:
+        for name in [n for n in os.environ if n.startswith("GIT_")]:
+            monkeypatch.delenv(name)
+        yield
+
+
 @pytest.fixture
 def git_repo() -> Generator[GitRepo]:
     with tempfile.TemporaryDirectory() as tmpdir:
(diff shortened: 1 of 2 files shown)


wkentaro:

Verdict: **recommend merge**.

Standards review found no blocking violations. The only optional nit is that `KEEP` could use a more specific name. Spec review found no missing, extra, or incorrect behavior. Local verification passed `make lint`, `git diff --check`, and all 358 tests. All seven GitHub checks pass across Linux, macOS, Windows, Python 3.10, and Python 3.14.

PR #199, opened 2026-08-06
fix: make file-level changes safe and atomic

### LINKED ISSUE #104: Conflicted files silently dropped: parser ignores combined-diff (diff --cc) output during merge/rebase
> *This was generated by AI during a kaizen pass.*

## Summary

During an unresolved merge/rebase conflict, `git diff` emits **combined-diff**
format (`diff --cc <path>`, hunk headers `@@@ ... @@@`, no `a/`…`b/` prefixes).
git-hunk's parser only understands the `diff --git a/… b/…` form, so conflicted
files are **silently dropped**: `list` reports nothing, exit code `0`, no warning,
even though the working tree has unmerged content.

## Reproduction

```sh
git init repo && cd repo
printf 'a\nb\nc\n' > f.txt && git add f.txt && git commit -m init
git switch -c other && printf 'a\nX\nc\n' > f.txt && git commit -am other
git switch -                     # back to the default branch
printf 'a\nY\nc\n' > f.txt && git commit -am main
git merge other                  # -> CONFLICT (content) in f.txt

git status --short               # UU f.txt
git diff --stat                  # f.txt | Unmerged  (git sees the change)

git-hunk list                    # => "No hunks."
git-hunk list --json             # => {"schema_version": 2, "hunks": []}
```

`git diff` reports f.txt as changed; git-hunk reports nothing.

## Root cause

- `split_file_diffs` (git_hunk/_hunk.py:125) splits on `(?=^diff --git )`, so a
  `diff --cc` block is never separated into its own file diff.
- `extract_file_path` (git_hunk/_hunk.py:158) only matches `diff --git a/… b/…`
  (and its quoted variant), returning `None` for a `diff --cc` header.
- Net effect in `parse_diff` (git_hunk/_hunk.py:240): the conflicted file
  produces no `Hunk`, so it vanishes with no error.

## Why this needs a decision (not a drive-by fix)

There is more than one reasonable answer, and each has a different blast radius on
the parser and the ADR 0001 Hunk model:

1. **Out of scope / documented no-op** — partial staging mid-conflict is an
   unusual workflow; git itself refuses some conflict operations. Maybe git-hunk
   should keep ignoring combined diffs but do so *loudly* (a warning to stderr, or
   a non-zero exit hint) rather than silently.
2. **Surface as a whole-file hunk** — emit a synthetic hunk (like binary/mode-only
   changes) so the file at least appears in `list`, without attempting to address
   the `@@@` body. Smaller change; `change_kind` would need a conflict/unmerged
   marker the model does not currently have.
3. **Full combined-diff support** — parse the `@@@ … @@@` two-column body so hunks
   are addressable during a conflict. Largest change; the per-line `op`/numbering
   model in ADR 0001 assumes single-parent diffs and would need extending.

This touches the `@@`-parsing core and potentially the `--json` schema
(`change_kind`, `lines[]`), so it wants a maintainer call before any code.

## Affected files

- `git_hunk/_hunk.py` — `split_file_diffs`, `extract_file_path`, `parse_diff`
- (potentially) `git_hunk/_git.py` — `get_diff`, if the diff invocation should
  change for the unmerged case
- ADR 0001 / `CONTEXT.md` — if `change_kind` or the Hunk model grows a conflict state

## Suggested minimal first step

If option 1 is chosen, the smallest safe change is to detect a `diff --cc` /
`diff --combined` block and emit a clear warning instead of silently producing
zero hunks, deferring real support to a follow-up. But that is a product call,
hence this triage.

### LINKED ISSUE #148: Empty staged new/deleted file is invisible to git-hunk (false clean-tree signal; unactionable commit/unstage trap)
> *This was generated by AI during a kaizen pass.*

## Summary

A staged (or committed-delta) **empty new/deleted file** is invisible to every `git-hunk` view. `git-hunk list` reports a clean tree while `git status` shows a real staged change, and the file cannot be managed through `git-hunk` at all. When such a file is the only thing staged, it silently blocks `git-hunk commit` with a tip that does not work.

## Reproduction (confirmed against `origin/main` @ 3c087be)

```console
$ git init -q repro && cd repro
$ printf 'x\n' > seed.txt && git add -A && git commit -qm init

$ touch empty.txt && git add empty.txt
$ git status --short
A  empty.txt                         # git: a staged file

$ git-hunk list
No hunks.                            # git-hunk: false "clean" signal

$ git-hunk list --staged --json
{ "schema_version": 2, "hunks": [] }

$ git-hunk unstage empty.txt
error: no changed file matches 'empty.txt'   # cannot manage it either
```

The commit trap, when the invisible file is the only staged change:

```console
$ printf 'x\ny\n' > seed.txt           # a normal unstaged change
$ git-hunk commit <seed-hunk-id> -m "..."
error: cannot commit: changes are already staged
  tip: commit them with 'git commit', or unstage with 'git-hunk unstage'
```

The tip is unactionable: `git-hunk unstage empty.txt` fails (the file was never registered as a hunk), so the user is forced out to raw `git` with no hint why.

## Root cause

`parse_diff` only emits a whole-file hunk for a bodyless diff when it is a chmod (`git_hunk/_hunk.py:290-305`):

```python
if len(parts) == 1:
    # ... An empty new/deleted file (A/D with no @@ body) is left out,
    # matching git-hunk's prior behavior.
    if change_kind == "M" and a_mode != b_mode:
        hunks.append(_whole_file_hunk(...))
    i += 1
    continue
```

An empty add/delete has a `diff --git` header, no `@@` body, and `change_kind` `A`/`D`, so it never becomes a `Hunk`.

## Why this needs a maintainer decision (not a blind fix)

The exclusion is **intentional and locked in** by `tests/e2e/list_test.py:192` (`test_empty_new_file_is_not_a_bogus_mode_hunk`), whose stated rationale was to avoid rendering a nonsensical `Mode None -> None` label. That test guarded the *rendering* problem; this issue is the distinct *visibility/trust* consequence. Any real fix changes that locked-in assertion and requires deciding:

1. **Should it surface at all, and as what?** e.g. a new whole-file kind like `Empty file (added)` / `Empty file (deleted)`, mirroring the existing `Binary file (added)` labels in `git_hunk/_ui.py`.
2. **Should `stage`/`unstage`/`discard`/`commit <path>` then manage these files** (same category question as untracked files, which are deliberately excluded from `stage <path>` per `tests/e2e/binary_test.py`)?
3. **`-l`/line-selection behavior** — reject it the same way binary/mode hunks already do?
4. Alternatively, is the intended contract simply "git-hunk operates on content hunks; empty adds/deletes are out of scope" — in which case the fix is only to make the `commit` block's tip actionable (or its message accurate) rather than to surface the file.

Options 1-3 and option 4 are meaningfully different products, which is why this is escalated rather than shipped.

## Affected files

- `git_hunk/_hunk.py:290-305` — `parse_diff` bodyless-diff branch.
- `tests/e2e/list_test.py:192` — the locked-in assertion any fix must revisit.
- `git_hunk/_ui.py` — where a new whole-file label would live, if option 1.
- `git_hunk/_cli.py` — `commit` staged-change guard / tip, if option 4.

### LINKED ISSUE #153: Rename with " b/" in old path mis-parses file field (unquoted diff --git header is ambiguous)
> *This was generated by AI during a kaizen pass.*

## Summary

`extract_file_path` (`git_hunk/_hunk.py:158`) parses the file path out of the
`diff --git a/<old> b/<new>` header line. For a **rename/copy whose old path
contains the literal substring `" b/"`** (a space followed by `b/`, no quoting),
the fallback regex splits at the wrong `" b/"` and returns a corrupted path.

## Reproduction (real git, end-to-end)

```sh
git init repo && cd repo
mkdir "x b"
printf 'l1\nl2\nl3\nl4\nl5\nl6\n' > "x b/y.txt"
git add -A && git commit -m init
git mv "x b/y.txt" z.txt
printf 'l1\nl2\nL3\nl4\nl5\nl6\n' > z.txt   # content change so it is not a pure rename (#53)
git add -A
```

git's header for this change is:

```
diff --git a/x b/y.txt b/z.txt
```

Parsing it:

```python
from git_hunk._hunk import parse_diff
from git_hunk._git import run_git
parse_diff(run_git('diff', '--cached'))[0].file
# => 'y.txt b/z.txt'   (expected 'z.txt')
```

`git status` correctly shows `R  "x b/y.txt" -> z.txt`.

## Why it happens

The unquoted `diff --git a/<old> b/<new>` header line is **genuinely ambiguous**
when a path contains a space: there is no way to recover the `a/`..`b/` boundary
from that line alone. `extract_file_path` handles two unambiguous cases first
(quoted paths; and identical old==new via a backreference), then falls back to:

```python
m = re.match(r"diff --git a/(.*?) b/(.*)", first_line)   # _hunk.py:171
return m.group(2) if m else None
```

The non-greedy `(.*?)` stops at the **first** `" b/"`, which lands inside the old
path (`x b/y.txt`) instead of at the true boundary, yielding `'y.txt b/z.txt'`.

## Impact

`Hunk.file` is the key every consumer uses:

- `list --json` / `show --json` report a path that does not exist in the repo.
- `_select_hunks` matches CLI path args against `hunk.file`, so
  `git-hunk stage/unstage/discard z.txt` (the file's real path per `git status`)
  fails with `no changed file matches 'z.txt'`. The hunk is reachable only by its
  content-hash id, and even then the reported `file` stays wrong, silently
  misleading any script/agent keying off it.

## Why this needs a decision (not a drive-by fix)

The header line is unparseable in isolation; a correct fix means changing the
**path-extraction strategy**, which is core code feeding every consumer and
overlaps open work:

1. **Read the authoritative line instead of the `diff --git` line.** For a rename
   git emits unambiguous `rename from <old>` / `rename to <new>` lines (and
   `--- a/<old>` / `+++ b/<new>` for content changes) that carry the path to
   end-of-line. This is the robust fix but re-shapes `extract_file_path` (which
   today intentionally reads only `first_line`).
2. **Use git's machine format** (`-z` / `--name-status` / `--raw`) as the source
   of truth for paths, keeping the textual diff only for hunk bodies. Bigger
   change; also the cleanest long-term answer.
3. **Tighten the fallback regex** (e.g. anchor on the `rename to`/`copy to` path).
   Smallest diff, but still first-line-only and easy to get subtly wrong.

This also brushes against #53 (surface renames), #40 (structured change-kind/mode
fields), and #67 (v2 `--json` schema), so the right path may be to fold it into
that direction rather than patch the regex in isolation.

## Affected files

- `git_hunk/_hunk.py` — `extract_file_path` (line 158, fallback at line 171),
  consumed by `parse_diff`.
- `git_hunk/_cli.py` — `_select_hunks` path matching against `hunk.file`.

Rare in practice (needs `" b/"` literally inside a renamed path) but it is silent
data corruption of the file field with no error, so parking for a maintainer call
on which strategy to adopt.

### LINKED ISSUE #154: Quoted-path renames drop the whole diff block, silently losing staged content changes
> *This was generated by AI during a kaizen pass.*

## Summary

When a **renamed** file has a path that git C-quotes (contains a tab, newline, backslash, or double-quote), `git-hunk` drops the **entire** file-diff block, **including any staged content changes** in that rename. `list`/`show`/`stage`/`unstage`/`discard`/`commit` all report a clean/empty result while a real staged content change exists, so an agent relying on git-hunk as its view of the working tree silently loses that change.

This is a **parity gap**: the fully-unquoted equivalent surfaces the content hunk correctly today.

## Root cause

`extract_file_path` (`git_hunk/_hunk.py:158-172`) only recognizes three header shapes:

```python
m = re.match(r'diff --git "a/(.+)" "b/\1"$', first_line)   # both quoted, old == new
...
m = re.match(r"diff --git a/(.+) b/\1$", first_line)        # both plain, old == new
...
m = re.match(r"diff --git a/(.*?) b/(.*)", first_line)      # both plain, any (returns new/group 2)
```

A rename header has `old != new`, so the two `\1`-backreference branches never match it. The third (fallback) branch matches only when the header starts with an unquoted `a/...` and contains a literal ` b/`. A quoted rename header does neither, so `extract_file_path` returns `None`, and `parse_diff` (`git_hunk/_hunk.py:250-253`) then skips the whole block — content hunk and all.

Because git quotes each half of the header independently, all three quoting-mix variants of a rename are dropped:

| header (real git output)                              | git-hunk hunks |
|-------------------------------------------------------|:--------------:|
| `diff --git "a/q\ta.txt" "b/q\tb.txt"` (both quoted)  | 0 (dropped)    |
| `diff --git "a/src\tq.txt" b/dstplain.txt`            | 0 (dropped)    |
| `diff --git a/srcplain2.txt "b/dst\tq.txt"`           | 0 (dropped)    |
| `diff --git a/oldname.txt b/newname.txt` (both plain) | 1 (correct)    |

## Reproduction (git 2.43.0)

```console
$ git init -q repo && cd repo
$ printf 'l1\nl2\nl3\nl4\nl5\nl6\n' > $'old\tname.txt'
$ git add -A && git commit -qm base
$ git mv $'old\tname.txt' $'new\tname.txt'
$ printf 'l1\nX2\nl3\nl4\nl5\nl6\n' > $'new\tname.txt'   # rename + a real content edit
$ git add -A

$ git -c core.quotePath=false diff --cached -U3 | head -3
diff --git "a/old\tname.txt" "b/new\tname.txt"
similarity index 83%
rename from "old\tname.txt"

$ git-hunk list --staged --json
{"schema_version": 2, "hunks": []}      # <- the staged X2 edit has vanished
```

Swap the tab out of the filenames (so no quoting) and the same edit surfaces as one `M` hunk under `new/name.txt`.

## Why this is not an existing issue

- **#53** (pure renames silently dropped) is about renames with **no** content hunk; even a full fix here would not synthesize a hunk for a 100%-similarity rename. Its own description claims "a rename that also changes content is partially handled: the content hunk lists under the destination path" — that is true for unquoted paths but **false** for quoted ones (this issue).
- **#153** (unquoted `" b/"` in old path mis-parses) is about the fallback regex returning a **wrong** path (mis-labeling), not returning `None` and dropping the block. Different code path, strictly milder failure mode (mislabel vs. total data loss).
- **#148** (empty staged new/deleted file invisible) is a different mechanism.

## Why this needs triage rather than a direct PR

The bug is clear, but the **fix direction is a maintainer call that overlaps #153**, which is already debating reworking this exact function:

1. **Add quoting-aware regex branches** for the three rename shapes (unquote the `b/` half). Small and local, but grows an already-branchy function and leaves the underlying header-parsing approach that #153 wants to revisit.
2. **Parse the `rename to <path>` body line** (or `git diff --raw` / `-z --name-status`) instead of re-deriving the path from the `diff --git` line. This is exactly the direction floated in #153 and would resolve #153, part of #53, and this issue together — but it is a larger change to `extract_file_path` / `parse_diff` and their tests.

Choosing between a surgical patch now vs. folding this into the #153 rework is the decision to make. There is also a scoping question: fix only rename-with-content (restore parity) or also handle 100%-similarity pure renames (converges with #53).

## Affected files

- `git_hunk/_hunk.py` — `extract_file_path` (lines 158-172), `parse_diff` block-skip (lines 250-253)
- No test currently exercises a rename at all (`grep -rn rename tests/` is empty); `tests/e2e/quoted_paths_test.py` covers quoted paths only for in-place modifications.

### LINKED ISSUE #161: Mode change leaks into every hunk selection: stage/commit/discard apply an unrequested chmod
> *This was generated by AI during a kaizen pass.*

## Summary

When a file has **both** a mode change (e.g. `chmod +x`) **and** one or more text hunks,
the mode change is not a selectable unit. It is silently attached to *every* hunk of that
file, so selecting any single hunk also applies the mode change. This reaches real commits
and silently reverts working-tree permission bits on `discard`.

This contradicts the documented guarantee in `README.md:122`:

> `commit` aborts if anything is already staged, so the commit contains exactly
> the selected hunks

## Reproduction

```sh
mkdir -p /tmp/modeleak && cd /tmp/modeleak && git init -q .
seq 1 30 > f.txt && chmod 644 f.txt && git add f.txt && git commit -qm init
python3 - <<'PY'
lines = [str(i) for i in range(1, 31)]
lines[1] = "CHANGED2"; lines[27] = "CHANGED28"
open("f.txt", "w").write("\n".join(lines) + "\n")
PY
chmod 755 f.txt          # unrelated exec-bit flip, far from either edit
git-hunk list
```

`list` shows two hunks and no mode unit:

```
unstaged:
f.txt
  0be3b10  @@ -1,5 +1,5 @@  +1 -1
  e746b86  @@ -25,6 +25,6 @@  +1 -1
```

### 1. `stage` leaks the mode change

```sh
git-hunk stage 0be3b10 && git diff --cached | head -3
```

Observed (the mode change is staged; only hunk `0be3b10` was selected):

```
diff --git a/f.txt b/f.txt
old mode 100644
new mode 100755
```

Selecting the *other* hunk instead leaks it identically.

### 2. `commit` bakes it into history

```sh
git-hunk commit 0be3b10 -m "only first hunk" && git log -1 -p | head -11
```

Observed: the commit's diff carries `old mode 100644` / `new mode 100755`.

### 3. `discard` reverts a chmod the user never touched

```sh
# with the same unstaged setup
stat -c '%a' f.txt   # 755
git-hunk discard 0be3b10
stat -c '%a' f.txt   # 664  <-- exec bit silently reverted
```

## Expected

Selecting one hunk id should affect that hunk's content only. For comparison, `git add -p`
on the exact same repo state offers the mode change as its own approvable unit:

```
(1/3) Stage mode change [y,n,q,a,d,j,J,g,/,?]?
(2/3) Stage this hunk [...]  @@ -1,5 +1,5 @@
(3/3) Stage this hunk [...]  @@ -25,6 +25,6 @@
```

git-hunk exposes only units 2 and 3, and applies unit 1 along with whichever of them is chosen.

## Root cause

`_patch.py:8-17` (`_extract_file_headers`) captures everything before the first `@@` line,
which includes `old mode` / `new mode`, and `build_patch` (`_patch.py:20-34`, line 32)
unconditionally prepends that header block to whatever subset of the file's hunks was
selected:

```python
patches.append(headers[filepath] + hunk_diffs + "\n")
```

Reached from `_apply_selection` (`_cli.py:264-267`), the path shared by
`stage` / `unstage` / `discard` / `commit`.

The parser already models the mode per hunk (`a_mode` / `b_mode`) and emits a dedicated
whole-file mode-only hunk when a file has *no* text hunks (`_hunk.py:290-303`). That
selectable unit simply does not exist when the file also has text hunks. PR #128 pins the
current parser behavior (mode metadata rides along on the content hunk) but does not touch
the apply path, so the leak is unaddressed.

## Why this needs a maintainer decision

There is more than one reasonable answer, and the blast radius reaches the documented
`--json` inventory that ADR 0001 governs:

1. **Emit a separate whole-file mode hunk even when text hunks exist**, and include the
   mode headers in the built patch only when that hunk is selected. Matches `git add -p`.
   Cost: changes the hunk inventory (`list`, `list --json`) for every chmod+edit file, so
   it is a visible output/schema change, and it needs an id scheme for that unit.
2. **Strip the mode headers from the patch unless a mode hunk is selected.** Fixes the
   leak, but on its own leaves *no* way to ever stage a mode change for a file that also
   has text edits, which is a functional regression.
3. **Attach the mode change to the first hunk only.** Smaller diff, still arbitrary and
   surprising.
4. **Declare it intended** and document the coupling in README/SKILL.md instead.

Option 1 looks right to me, but it changes documented output shape and overlaps #40's
modeling territory, so I did not want to pick it unattended.

## Affected files

- `git_hunk/_patch.py` (`_extract_file_headers`, `build_patch`)
- `git_hunk/_hunk.py` (whole-file mode hunk construction)
- `git_hunk/_cli.py` (`_apply_selection`)
- `README.md:122`, `git_hunk/skills/core/SKILL.md:43` (the "exactly the selected hunks" guarantee)

Verified against `origin/main` @ 3c087be.

### LINKED ISSUE #189: Make file-level edge cases safe and atomic
> *This was generated by AI.*

## Agent brief

Implement the v0.3.0 file-state safety contract after Repository path semantics lands.

## Decision and plan

- [File-level safety decision](https://github.com/wkentaro/git-hunk/issues/182#issuecomment-5199677332)
- [Exact release checklist](https://github.com/wkentaro/git-hunk/blob/45758d791edbb207dc1682df04aaf45af440ce5d/docs/research/186-v0.3.0-release-candidate-checklist.md#3-make-file-level-edge-cases-safe-and-atomic)

## Scope

Resolve these source issues:

- [Conflicted files are silently dropped](https://github.com/wkentaro/git-hunk/issues/104)
- [Empty staged new or deleted file is invisible](https://github.com/wkentaro/git-hunk/issues/148)
- [Rename with b/ in the old path misparses the file field](https://github.com/wkentaro/git-hunk/issues/153)
- [Quoted-path renames drop the whole diff block](https://github.com/wkentaro/git-hunk/issues/154)
- [Mode change leaks into every Hunk selection](https://github.com/wkentaro/git-hunk/issues/161)

Reject detected rename, copy, and unmerged states before output or mutation. Surface mode and text changes as separate Hunks. Surface empty tracked additions and deletions as whole-file Hunks. Validate mixed selections before the first mutation.

Full rename support remains in [Surface renamed files in list](https://github.com/wkentaro/git-hunk/issues/53).

## Acceptance

- Cover quoted and ambiguous paths, copies, all unmerged stage sets, empty files, independent mode and text operations, mixed selections, dry runs, and arbitrary path bytes.
- Never return partial inventory, false clean success, or partial mutation.
- Update affected public docs, bundled skills, and the Unreleased changelog with this pull request number.
- The pull request must close the five listed source issues and leave full rename support open.
- Run focused tests, full tests, and full lint.

wkentaro:
> *This was generated by AI.*

## Summary

- reject rename, copy, and unmerged index states before output or mutation
- represent mode changes and text edits as independent selectable Hunks
- represent empty tracked additions and deletions as selectable whole-file Hunks
- validate mixed selections and dry runs before the first mutation
- document the supported file-state model for users and agents

## Why

Git reports these file states with different diff and index shapes. Treating all of
them as ordinary text hunks could hide changes or apply only part of a selection.

## Test plan

- [x] added end-to-end coverage for rename, copy, all unmerged stage sets, empty files, mode and text combinations, dry runs, quoted paths, and arbitrary path bytes
- [x] `make test` (`614 passed`)
- [x] `make lint`

Closes #189
Closes #104
Closes #148
Closes #153
Closes #154
Closes #161

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -15,6 +15,11 @@ repository. You can run the suite from inside `git rebase --exec`, a hook,
 Keep this behavior. `tests/git_env_test.py` verifies that the suite does not
 change the outer repository.
 
+Pass Git plumbing protocols through binary subprocess input. Text-mode stdin
+converts LF to CRLF on Windows and can make commands such as
+`git update-index --index-info` silently ignore records. Mark only filename
+cases that Windows does not allow as skipped, not the full behavior test.
+
 ## Agent skills
 
 ### Issue tracker
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -69,6 +69,13 @@ and this project adheres to
 
 ### Fixed
 
+- Reject detected renames, copies, and unmerged index entries before showing a
+  partial inventory or applying a selection. These states now fail without
+  changing the index or working tree (#199).
+- Treat file mode changes and text edits as independent Hunks, and list empty
+  tracked additions and deletions as whole-file Hunks. Each change can now be
+  selected independently, and mixed selections validate before mutation
+  (#199).
 - Make partial line selection safe and bounded: reject ambiguous grouped
   replacements atomically; validate range endpoints before expansion; preserve
   no-newline state by patch side; and reject submodule pointer line selection
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -7,23 +7,25 @@
   one exact Repository path. They do not expand directories, globs, or Git
   pathspec syntax.
 
-- **Hunk** — the atomic, addressable unit of git-hunk. A contiguous change a user can
-  `stage` / `show` / `discard` by `id`. Usually one `@@` section of a unified diff; for
-  binary, mode-only, or type changes it is a synthetic **whole-file hunk** (no `@@` range).
-  The top-level object in `--json` (the tool is hunk-centric, not file-centric).
+- **Hunk**: the atomic, addressable unit of git-hunk. A contiguous change a user can
+  `stage` / `show` / `discard` by `id`. Usually one `@@` section of a unified diff. A
+  binary, mode-only, type, or empty tracked file change is a synthetic **whole-file
+  hunk** with no `@@` range. It is the top-level object in `--json` because the tool is
+  hunk-centric, not file-centric.
 
 - **Change group**: a maximal run of changed `-` and `+` lines with no context line
   between them. Partial line selection is unrestricted for pure additions, pure
   deletions, and one-for-one replacements. A larger grouped replacement is selected
   as a whole or not selected.
 
-- **Whole-file hunk** — a hunk with no `@@` text range: a binary change, a mode-only
-  (chmod) change, or a type change (e.g. file ↔ symlink). Has `header: null` and (in
-  `show --json`) `lines: []`.
+- **Whole-file hunk**: a hunk with no `@@` text range: a binary change, a mode-only
+  (chmod) change, a type change (e.g. file to symlink), or an empty tracked addition
+  or deletion. Has `header: null` and, in `show --json`, `lines: []`.
 
-- **change_kind** — the git status letter for the hunk's file: `A` added, `D` deleted,
-  `M` modified, `T` typechange. Always present. `R` (rename) / `C` (copy) are reserved,
-  not yet produced (see #53). Mirrors `git diff --name-status`.
+- **change_kind**: the git status letter for the hunk's file: `A` added, `D` deleted,
+  `M` modified, `T` typechange. Always present. `R` (rename) and `C` (copy) are
+  reserved and currently rejected before inventory or mutation (see #53). Mirrors
+  `git diff --name-status`.
 
 - **a_mode / b_mode** — the file's git mode (6-digit octal *string*, e.g. `"100644"`) on
   the pre-image (`a`) and post-image (`b`) side; `null` when that side does not exist.
@@ -68,3 +70,6 @@
   free text.
 - The internal patch text fed to `git apply` preserves git's bytes verbatim; only the
   *JSON projection* is normalized (bared header, byte-safe union).
+- A mode change and text changes in one file are separate, independently selectable
+  Hunks.
+- Detected rename, copy, and unmerged states fail before inventory output or mutation.
--- a/README.md
+++ b/README.md
@@ -99,6 +99,14 @@ shell passes them unchanged. For example, from `sub/`, `same.txt` selects the
 file at the worktree root, while `sub/same.txt` selects the file inside `sub/`.
 `show` remains ID-only.
 
+### Unsupported repository states
+
+git-hunk rejects detected rename, copy, and unmerged index states before it
+writes inventory output or changes the repository. This prevents partial JSON,
+partial inventory, false clean results, and partial mutation. Resolve an
+unmerged index with Git before retrying. Full rename and copy support is not yet
+available; it remains tracked in [#53](https://github.com/wkentaro/git-hunk/issues/53).
+
 ### List hunks
 
 ```bash
@@ -142,8 +150,14 @@ Line selection accepts any subset of a pure addition, pure deletion, or
 one-for-one replacement. A grouped replacement with multiple deleted or added
 lines must be selected as a whole or not selected. Numeric range endpoints are
 checked against the Hunk before expansion, and no-newline state is preserved for
-each patch side. Submodule pointer changes do not support line selection. Select
-the Hunk as a whole.
+each patch side. Submodule pointer changes and whole-file Hunks do not support
+line selection. Select the Hunk as a whole.
+
+A binary, mode-only, type, or empty tracked file change is a whole-file Hunk.
+Plain output labels empty tracked changes as `Empty file (added)` or
+`Empty file (deleted)`. When one file has a mode change and text edits, the mode
+and each text range are separate Hunks. Selecting text does not apply the mode
+change, and selecting the mode Hunk does not apply text.
 
 ### Commit
 
@@ -199,11 +213,11 @@ body; `show --json` adds a structured `lines` array. A `show --json` hunk
 | `id`             | string         | Stable, content-based hunk id (7-char SHA-256 prefix); accepts prefixes; empty for an `untracked` entry, which no command can address. |
 | `file`           | union          | Repository path of the changed file, as a byte-safe `{text\|bytes}` union (see below).                                                 |
 | `status`         | string         | One of `staged`, `unstaged`, `untracked`.                                                                                              |
-| `change_kind`    | string         | Git status letter: `A` added, `D` deleted, `M` modified, `T` typechange (`R`/`C` reserved). Always present.                            |
+| `change_kind`    | string         | Git status letter: `A` added, `D` deleted, `M` modified, `T` typechange (`R`/`C` reserved and currently rejected). Always present.     |
 | `a_mode`         | string \| null | 6-digit octal git mode on the pre-image side; `null` when that side does not exist.                                                    |
 | `b_mode`         | string \| null | 6-digit octal git mode on the post-image side; `null` when that side does not exist.                                                   |
 | `binary`         | bool           | Whether the change is binary. Always present.                                                                                          |
-| `header`         | string \| null | The hunk's bare `@@ -a,b +c,d @@` range; `null` for a whole-file (binary, mode-only, or type) change.                                  |
+| `header`         | string \| null | The hunk's bare `@@ -a,b +c,d @@` range; `null` for a whole-file (binary, mode-only, type, or empty tracked file) change.              |
 | `context_before` | union \| null  | The function/section git names after the `@@` header, as a `{text\|bytes}` union; `null` when there is none.                           |
 | `additions`      | int            | Number of added lines.                                                                                                                 |
 | `deletions`      | int            | Number of removed lines.                                                                                                               |
@@ -237,10 +251,11 @@ renaming, removing, or changing the type of an existing field bumps it. (Before
 
 ## How it works
 
-1. Parses `git diff` output into individual hunks
-2. Assigns each hunk a stable, content-based ID (SHA-256 prefix)
-3. For staging: reconstructs a minimal patch and pipes it through `git apply --cached`
-4. For discarding: reconstructs a reverse patch and applies it to the working tree
+1. Rejects detected rename, copy, and unmerged states.
+2. Parses `git diff` output into individual hunks.
+3. Assigns each hunk a stable, content-based ID (SHA-256 prefix).
+4. For staging, reconstructs a minimal patch and pipes it through `git apply --cached`.
+5. For discarding, reconstructs a reverse patch and applies it to the working tree.
 
 IDs are derived from the changed lines, not the `@@` line numbers that shift as
 you stage other hunks -- so an unrelated hunk keeps its id, while staging only
--- a/docs/adr/0001-json-v2-hunk-schema.md
+++ b/docs/adr/0001-json-v2-hunk-schema.md
@@ -42,8 +42,9 @@ values derive from one source and cannot diverge.
 
 `change_kind`: one of `"A"` (added), `"D"` (deleted), `"M"` (modified), `"T"` (typechange).
 **Always present, non-null.** `"R"` (rename) and `"C"` (copy) are **reserved** for when
-renames land (#53). Single letters match `git diff --name-status` and are consistent with
-the per-line `op` field's git-native single chars (see §6).
+rename and copy support lands (#53). These states are currently rejected before
+inventory output or mutation. Single letters match `git diff --name-status` and are
+consistent with the per-line `op` field's git-native single chars (see §6).
 
 ### 3. `a_mode` / `b_mode` — octal strings, always present
 
@@ -73,7 +74,8 @@ mode-only one (both have empty bodies). Do not infer binary-ness from an empty b
 
 `header`: for a text hunk, the **bare** `@@ -a,b +c,d @@` range with git's trailing
 section heading stripped (#50). The section heading lives only in `context_before`.
-For a whole-file hunk (binary, mode-only, or type change — no `@@` range), `header` is `null`.
+For a whole-file hunk (binary, mode-only, type, or empty tracked file change with no
+`@@` range), `header` is `null`.
 
 The internal `Hunk.diff` attribute used to build patches for `git apply` keeps git's
 original `@@` line **verbatim** (including the section heading); only the JSON `header`
@@ -194,6 +196,7 @@ Binary modify (whole-file hunk):
 
 ## Out of scope
 
-- Renames/copies (#53) — `R`/`C` are reserved in `change_kind` but not produced.
+- Rename and copy support (#53). `R` and `C` are reserved in `change_kind`, not
+  produced, and currently rejected before output or mutation.
 - Hunk-id stability and untracked/new-file staging (#13, #22).
 - The schema-doc/versioning *process* (#23); this ADR is the schema, not the doc tooling.
--- a/docs/adr/0002-repository-path.md
+++ b/docs/adr/0002-repository-path.md
@@ -111,8 +111,8 @@ no path operand.
 
 ## Out of scope
 
-- Mid-mutation atomicity. A selection is validated completely before the first
-  mutation, but a failure between the textual `git apply` leg and the whole-file leg
-  can still leave the first applied. That is decided with the other file-level
-  edge cases, not here.
-- Rename and copy states, which remain rejected (#53).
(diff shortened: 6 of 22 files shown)


wkentaro:

**Verdict: recommend-merge**

Two-axis review (Standards, Spec) against #189 and the five source issues, with empirical checks on the branch.

## Spec

- The rename/copy/unmerged rejection runs before any inventory output or mutation. `_get_hunks` is the single gate; `list`, `show`, `stage`, `unstage`, `discard`, and `commit` all route through it.
- Mode and text changes are independent Hunks with correct header handling in `build_patch`. Empty tracked additions and deletions surface as `Empty file (added|deleted)` whole-file Hunks, which also fixes the unactionable `commit` tip from #148.
- Mixed selections validate everything Git can validate before the first mutation (patch `--check`, `add --dry-run`, `rm --dry-run`). The reverse legs use `git restore`, which has no dry-run flag, so the residual cross-process race is documented in ADR 0002 rather than eliminated. This is the achievable maximum without snapshotting the index.
- Test matrix covers quoted and ambiguous paths, copies, all unmerged stage sets, empty files, mode and text combinations, dry runs, and non-UTF-8 path bytes. 614 tests pass; CI is green on all platforms.
- Docs, both bundled skills, and the Unreleased changelog carry #199. #53 stays open as required.

## Standards

No hard violations. Minor judgement calls, none blocking:

- `tests/e2e/conftest.py`: `run_ok` now returns `r.stdout or r.stderr`; the name promises stdout, and the fallback lets an assertion match error text on a successful run.
- `_apply_selection` introduces a patch-leg vs file-command-leg distinction that CONTEXT.md's glossary does not name yet.
- The changelog lists the mode/text Hunk split under Fixed; it changes the inventory shape for chmod+edit files, so Changed is arguable.

## Follow-up filed (not blocking)

The copy guard pins `--find-copies-harder -l0`, so a plain `cp` of an unmodified file plus `git add` now blocks every git-hunk command, and exhaustive detection runs twice per invocation. This matches the acceptance contract ("A detected copy, with or without a text edit"), but the blast radius and cost deserve an explicit call: #200.

PR #207, opened 2026-08-07
feat: add make eval target and fix eval preflight pycache failure

### LINKED ISSUE #206: Eval preflight fails on __pycache__ because .gitignore does not cover it

Found while running the eval from PR #203 on a developer machine.

`resolve_environment()` runs `git status --porcelain=v1 --untracked-files=all` with `GIT_CONFIG_GLOBAL` set to `os.devnull`, so a global `core.excludesFile` that normally hides `__pycache__/` does not apply. The repo `.gitignore` has no `__pycache__` rule, so the preflight reports "the eval checkout must be clean" on any checkout where Python has run. Importing `eval` and `git_hunk` recreates `__pycache__/` before the status check runs, so deleting the directories alone does not help; the run needs `python -B`.

Fix: add `__pycache__/` to `.gitignore`. Consider documenting `python -B` or ignoring `*.pyc` as well.

Workaround:
```bash
find . -name __pycache__ -type d -prune -exec rm -rf {} +
uv run python -B -m eval
```

Does not affect the PR #203 verdict; the deterministic tests and CI are unaffected.

wkentaro:
> *This was generated by AI.*

Running the eval on a fresh machine failed before the model phase: the preflight saw `__pycache__` as untracked because only the global gitconfig ignored it (#206), and the Claude Code 2.1.222 pin required a manual `npm install` plus PATH setup. The `__pycache__/` gitignore rule landed on `main` directly (c794cba). This PR adds the `make eval` target: it reads the pin from `eval/config.py`, installs that version once under `~/.cache/git-hunk/`, and runs `python -m eval` with it on PATH. The recipe starts with `set -eu` so a failed pin read aborts instead of letting `npm install` fall back to an unpinned latest.

`make eval EVAL_ARGS="--task drop_debug_lines"` runs a single-task smoke; plain `make eval` runs the full qualifying set.

## Test plan

- [x] `make eval EVAL_ARGS=--help` resolves the pin, reuses the cached install under `~/.cache/git-hunk/claude-2.1.222`, and reaches the runner
- [x] pinned `claude --version` reports `2.1.222 (Claude Code)`
- [x] simulated pin-read failure aborts the recipe before `npm install` runs
- [x] `make lint` and `make test` (696 passed)

Closes #206

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,7 @@ ifneq ($(OS),Windows_NT)
 	SHELL := bash
 endif
 
-.PHONY: help setup format lint test coverage
+.PHONY: help setup format lint test coverage eval
 .DEFAULT_GOAL := help
 
 PYTEST_ARGS ?= --numprocesses=auto
@@ -39,3 +39,10 @@ test:  # Run tests
 
 coverage:  # Run tests with coverage
 	$(MAKE) test PYTEST_ARGS="--cov=git_hunk --cov-report=term-missing"
+
+eval:  # Run the agent eval with pinned Claude Code (spends real model usage)
+	@set -eu; \
+	pin=$$(uv run --no-sync python -c "from eval.config import CLAUDE_CODE_VERSION; print(CLAUDE_CODE_VERSION)"); \
+	prefix="$$HOME/.cache/git-hunk/claude-$$pin"; \
+	test -x "$$prefix/node_modules/.bin/claude" || npm install --prefix "$$prefix" "@anthropic-ai/claude-code@$$pin"; \
+	PATH="$$prefix/node_modules/.bin:$$PATH" uv run python -m eval $(EVAL_ARGS)


wkentaro:

**Verdict: recommend-merge** (head a11c0a3, all 7 checks green)

Two-axis review (standards + spec vs #206), then one fix pushed this tick:

- Added `set -eu` at the start of the `eval` recipe. Without it, a failed pin read (for example a fresh machine before `make setup`) left `$pin` empty and `npm install "@anthropic-ai/claude-code@$pin"` fell back to an unpinned install in a mispinned cache directory. Verified: a simulated pin-read failure now aborts before `npm install`; the happy path (`make eval EVAL_ARGS=--help`) reuses the cached 2.1.222 install and reaches the runner.
- The branch was already rebased onto `main`, so the `.gitignore` commit (identical to c794cba on `main`) is gone and the PR carries one commit. PR body updated to match.

Review notes that need no change:

- Spec: the `PATH` prepend reaches both the preflight (`claude --version`) and the solver; `make_subprocess_env()` strips only `GIT_*`, so the pinned binary resolves. With `__pycache__/` ignored, `python -B` from #206 is no longer needed.
- The recipe is bash-only, but `make eval` is not expected to work on Windows; no CHANGELOG entry since existing entries cover CLI-user-facing changes only.

PR #217, opened 2026-08-08
perf: streamline the git-hunk agent workflow

wkentaro:
Stacked on #216; merge that PR first. The bundled core skill previously encouraged repeated inventory, diff, and verification calls, inflating agent context and cost.

The paired eval kept 8/8 success while reducing git-hunk from 27 to 14 tool calls and from $0.3366 to $0.1670.

## Test plan

- [x] paired agent eval: 8/8 variants passed
- [x] `uv run pytest -q`: 697 passed
- [x] `uv run ty check`
- [x] `uv run ruff check .`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,8 @@ and this project adheres to
 
 ### Changed
 
+- Streamline the core skill around one inspection call and one execution call,
+  reducing redundant agent tool use (#217).
 - **Breaking:** Use durable Hunk identity from ADR 0003. JSON returns full
   SHA-256 IDs and `id_stability`; human output shows unique prefixes and marks
   Conditional IDs; commands accept unambiguous case-insensitive prefixes from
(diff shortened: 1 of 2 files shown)


wkentaro:
### Fable final verdict: APPROVE

Reviewed the complete two-file diff against current main, repository requirements, CLI behavior, checks, and PR discussion.

- The streamlined core skill remains accurate about staged-index guards, stable and Conditional IDs, exact repository paths, partial operations, untracked files, and destructive discard confirmation.
- The two-call workflow has appropriate escape hatches when information is missing or IDs become invalid.
- The changelog entry is correctly under Unreleased and references #217.
- All seven CI jobs pass across Linux, macOS, Windows, Python 3.10, and Python 3.14.
- The PR is cleanly mergeable with no unresolved review discussion.

Residual risk: this primarily changes agent guidance, so effectiveness still depends on model instruction-following; the explicit exceptions and safety boundaries make that non-blocking.

Recommendation: merge.

Issue #114, opened 2026-07-12
Partial line-selection silently corrupts content when the selection splits a grouped (-,-,+,+) change


wkentaro:
> *This was generated by AI during a kaizen pass.*

## Summary

Partial line-selection (`-l` / `--include-matching` / `--exclude-matching`) **silently corrupts file content** when the selection splits *inside* a grouped change — the `-…-…+…+…` shape git emits for two or more **adjacent** changed lines. Kept lines are rendered in raw diff order, which reorders content relative to the change. `git apply` accepts the malformed patch, so there is no error. Affects `stage`, `unstage`, `discard`, and `commit -l` alike (all route through `filter_hunk_lines`).

## Reproduction

```console
$ printf 'a\nb\nc\n' > f.txt && git add f.txt && git commit -qm init
$ printf 'a\nB\nC\n' > f.txt          # git diff body: " a  -b  -c  +B  +C"
$ git-hunk show <id>
  1  a
  2 -b
  3 -c
  4 +B
  5 +C
$ git-hunk stage <id> -l 2,4          # stage b->B only, leave c->C
$ git show :f.txt
a
c
B                                     # EXPECTED a\nB\nc\n  — c and B swapped
```

The reverse (unstage) direction corrupts identically:

```console
$ git-hunk unstage <id> -l 3,5        # unstage c->C only, keep b->B staged
$ git show :f.txt
a
c
B                                     # EXPECTED a\nB\nc\n
```

Both `stage`/`discard` (forward) and `unstage`/`discard` (reverse) are affected via distinct selected/kept combinations.

## Root cause

`git_hunk/_lines.py` — `_select_body_lines` (L66-91) walks the body in **raw diff order**, converting an unselected `-` line into context in place, and `_render_body_lines` (L94-116) emits that order verbatim. For `-b -c +B +C` with `-l 2,4`, the unselected `-c` becomes context ` c` *before* the kept `+B`, so the new side reads `a, c, B` instead of `a, B, c`.

Every existing test uses the **interleaved** shape (`tests/e2e/line_selection_test.py`, `tests/unit/_lines/filter_test.py`'s `_TWO_GROUP = " a -b +B  c -d +D"`), where each change is separated by context, so this case is entirely uncovered.

## Why this needs a maintainer decision (not a blind fix)

There is more than one internally-consistent semantic for "split a grouped change," and the current output is a coherent implementation of one of them:

[...]

PR #220, opened 2026-08-08
perf: cut the agent's tool calls and skill tokens

wkentaro:
## Summary

- Count the tool calls in each eval trace and report them in the run summary table, the usage line, and the run manifest. Tool calls are the metric the git-hunk-vs-bare-Git comparison is about, and the table carried only turns and cost.
- Let a partial-line or Conditional Hunk ID commit chain its path-addressed cleanup and the closing `git-hunk list` into one Bash call, instead of splitting them across two.
- Make that closing `git-hunk list` the whole verification, so the agent stops re-checking with `git log` / `git status`, and bound the closing report to one line per commit.
- Condense the core skill to the rules an agent acts on, and keep the Hunk ID nuance in the splitting section so a plan of complete hunks never pays to reason about it.
- Amend ADR 0003 and the README: a Repository path is ID-independent, so a fresh inventory is not the only way to address what remains after an ID break.

## Measured effect

Same four tasks, `claude-sonnet-5`, one run each.

Before, on `main` (4402671) — the table has no tool-call column yet:

| Task                      | git-hunk              | bare-git                     |
| ------------------------- | --------------------- | ---------------------------- |
| split_refactor_vs_feature | PASS · 4t · $0.05     | PASS · 5t · $0.02            |
| separate_mixed_hunks      | PASS · 7t · $0.06     | FAIL partition · 15t · $0.16 |
| drop_debug_lines          | PASS · 5t · $0.05     | PASS · 8t · $0.08            |
| protect_unrelated_work    | PASS · 4t · $0.04     | PASS · 4t · $0.02            |
| **total**                 | **4/4 · 20t · $0.20** | **3/4 · 32t · $0.28**        |

After, on this branch:

| Task                      | git-hunk                    | bare-git                    |
| ------------------------- | --------------------------- | --------------------------- |
| split_refactor_vs_feature | PASS · 3c · 4t · $0.04      | PASS · 5c · 6t · $0.02      |
| separate_mixed_hunks      | PASS · 3c · 4t · $0.04      | PASS · 23c · 24t · $0.21    |
| drop_debug_lines          | PASS · 3c · 4t · $0.04      | PASS · 8c · 9t · $0.07      |
| protect_unrelated_work    | PASS · 3c · 4t · $0.04      | PASS · 2c · 3t · $0.02      |
| **total**                 | **4/4 · 12c · 16t · $0.17** | **4/4 · 38c · 42t · $0.32** |

Every task now costs a flat 3 tool calls and 4 turns: one skill load, one inspection, one execution. That is the floor these tasks allow.

These are single paired runs, so treat the deltas as directional. Tool-call counts are stable across repeats; cost moves a few percent and still carries the prompt-cache ordering caveat the table prints.

## Note on the skill diff

Review caught two defects the condensation introduced, both fixed here and both reproduced against the CLI first:

- The condensed rule only barred a later *Hunk ID*, so it permitted a second `-l` after a partial-line commit. Body positions shift too, and `git-hunk commit a.py -l 4 -m x && git-hunk commit a.py -l 5 -m y` silently commits the wrong line rather than erroring.
- An intermediate wording claimed content matching survives an ID break. It does not: `--include-matching` / `--exclude-matching` need a single-hunk selection, and a bare path over a multi-hunk file exits 1 with `line selection requires exactly one hunk`.

## Test plan

- [x] `make test` (720 passed) and `make lint` pass on every commit in isolation, checked in a detached worktree
- [x] `tests/eval/model_test.py::test_trace_usage_counts_each_tool_use_once` covers the dedupe and the assistant-event filter
- [x] `tests/eval/summary_test.py` and `tests/eval/runner_test.py` pin the tool-call column in the rendered table
- [x] `make eval` passes 4/4 on the git-hunk variant with the numbers above

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,9 @@ and this project adheres to
   readable. The exit code now reports the subject under test (the git-hunk
   variant) and solver errors, so a graded bare-Git failure is evidence instead
   of a red run (#221).
+- Agent eval summary tables, usage lines, and run manifests count the tool calls
+  each task variant made, so the workflow comparison reports the metric it is
+  about (#220).
 - Agent evals run every task with git-hunk and bare Git from equivalent initial
   state for direct workflow and cost comparison, and `--help` lists the
   available tasks (#216).
@@ -45,6 +48,14 @@ and this project adheres to
 
 ### Changed
 
+- Ask for a one-line-per-commit close in the core skill, so the agent's final
+  report stops restating the diff it just committed (#220).
+- Condense the core skill to the rules an agent acts on, trimming the text it
+  loads before its first inspection call (#220).
+- Let a partial-line or Conditional Hunk ID commit chain path-addressed cleanup
+  and the closing `git-hunk list` into one call, and make that list the agent's
+  final check, so dropping a debug line out of a hunk costs one Bash call
+  instead of two (#220).
 - Streamline the core skill around one inspection call and one execution call,
   reducing redundant agent tool use (#217).
 - Condense the logical-commits skill around commit judgment and make core
--- a/README.md
+++ b/README.md
@@ -123,8 +123,10 @@ creates new Hunks with new IDs.
 Hunks with the same Repository path and patch content form a Duplicate Hunk
 group. Each member gets a unique Conditional Hunk ID, shown with a
 `conditional` label in human output and `"id_stability": "conditional"` in JSON.
-The ID can change when its Duplicate Hunk group changes. Run `git-hunk list`
-again after a partial operation or an operation on a Conditional Hunk ID.
+The ID can change when its Duplicate Hunk group changes. After a partial-line
+operation or an operation on a Conditional Hunk ID, address anything remaining
+by Repository path, which is ID-independent, or run `git-hunk list` again for
+the new IDs.
 
 ### List hunks
 
--- a/docs/adr/0003-durable-hunk-identity.md
+++ b/docs/adr/0003-durable-hunk-identity.md
@@ -16,9 +16,10 @@ staged and unstaged state.
 Partial-line operations create new Hunks with new IDs. Each member of a Duplicate
 Hunk group receives a unique Conditional Hunk ID. A Conditional Hunk ID can change
 when its group changes. Every JSON Hunk reports `id_stability` as `"stable"` or
-`"conditional"`. Human output marks each Conditional Hunk ID. The bundled skills
-must get a new inventory after a partial-line operation or an operation on a
-Conditional Hunk ID.
+`"conditional"`. Human output marks each Conditional Hunk ID. After a
+partial-line operation or an operation on a Conditional Hunk ID, the bundled
+skills must either get a new inventory or address what remains by Repository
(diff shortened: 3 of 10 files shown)


wkentaro:
## Final verdict (Fable 5): MERGE

An independent Fable 5 agent reviewed the branch after the five earlier review rounds, with instructions to judge fitness to merge rather than re-litigate nits. It verified the skill's rules empirically against the CLI in throwaway repositories.

> I empirically confirmed in throwaway repos that the failure class the earlier iteration had — silently committing wrong lines via a stale reference — is real (a stale Conditional Hunk ID committed the "three" duplicate under the message "two" with exit 0), and that the current SKILL.md wording forbids exactly the dangerous cases: after a partial-line or Conditional-ID operation, "only a bare Repository path may follow in that call," and paths are ID-independent by construction (`_cli.py:_select_hunks` matches path-first; IDs are content hashes in `_hunk.py`, so stale stable IDs fail loudly with "hunk not found" while only conditional ordinal-hashes can silently remap — the isolated case). Both documented chains work as written: `commit --exclude-matching ... && discard <path> && list` and conditional-ID commit followed by a path commit both produce exactly the intended commits. The eval change is a correct, honestly-tested derived metric (distinct tool-use IDs from assistant events, de-duped against re-streamed messages, reported-not-scored, matching ADR 0004), and the six commits are each self-contained with messages that match their content.

720 tests pass and lint is clean at the tip; every commit also passes both in isolation in a detached worktree.

### Outstanding, all non-blocking

- The relaxation that allows complete-hunk ID operations **before** a Conditional Hunk ID operation in one call (the old rule required the conditional op to be the *only* mutation) is sound — committing a disjoint hunk cannot reorder or re-member a duplicate group, since ordinals come from monotonic worktree positions — but nothing in the test suite pins it. It rests on the `assign_hunk_ids` design.
- The ID-validity rule moved from the core loop into "Splitting and cleanup". The core loop's guarantee is scoped to "a whole plan of complete hunks", so it is implicitly safe, but an agent skimming only the core loop no longer meets the conditional/partial caveat in step 2 itself.
- The measured improvement is one paired run per task. Directionally plausible, statistically weak — please don't cite it as more than that.
- Cosmetic: `format_usage` prints `1 tool calls` with no singularization, and the tests enshrine it. This matches the pre-existing `{turns} turns` style rather than introducing a new inconsistency.

### Correction: the `git-hunk stage -l` defect I reported does not exist

An earlier revision of this comment claimed `git-hunk stage <id> -l <lines>` accepts an
addition-only subset of a grouped replacement and emits a syntactically broken file.
**That is wrong, and I withdraw it.** Re-verified against this checkout's build
(`0.2.1.dev276`), the guard behaves correctly:

```console
$ git-hunk stage 210be83 -l 12-28
error: cannot partially select lines 11-50: grouped replacement (deletions: 1, additions: 39); select all or none
```

The broken output came from `git-hunk 0.2.1.dev53` at `~/.local/bin/git-hunk`, which is
first on `PATH` and predates the guard shipped for #114. `_validate_group` in
`git_hunk/_lines.py` already covers this case. No issue to file.

Issue #225, opened 2026-08-09
Content match hitting one side of a one-for-one replacement silently commits a broken half


wkentaro:
## Symptom

When `--include-matching` (or `-l`) selects only the `-` or only the `+` line of a one-for-one replacement pair, git-hunk commits that half without any warning: a deletion-only commit that removes the line outright, or an addition-only commit that duplicates it.

## Reproduction

```bash
# base: a file whose hunk replaces
#   -    return session.get(url)
#   +    return session.get(url, timeout=30)
git-hunk commit <id> --include-matching 'session.get(url)' -m "Add request timeout"
```

The pattern `session.get(url)` matches the `-` line but not the `+` line (`session.get(url, timeout=30)` has a comma after `url`). git-hunk reports `committed 1 hunk` and the commit contains only:

```diff
-    return session.get(url)
```

## Why it matters

The eval task `separate_formatter_noise` caught `claude-sonnet-5` doing exactly this on its first run: the fix commit deleted the `return` line, and the follow-up whole-path commit swept the `+` line into the formatting commit. The final tree was right, so nothing at the CLI surface ever signaled the broken intermediate commit.

Grouped replacements wider than one-for-one already get a guard (`cannot partially select lines N-M: grouped replacement ...; select all or none`). The one-for-one case has legitimate one-sided uses (committing a pure deletion), so an unconditional error is wrong, but the silent default is a footgun.

## Possible directions

- Warn (or require a flag) when a selection takes exactly one side of a one-for-one replacement pair.
- At minimum the bundled core skill now tells agents a replacement needs a pattern both lines contain; this issue tracks whether the CLI itself should guard it.

wkentaro:
> *This was generated by AI.*

## Decided scope (from #232, the v0.3 admit list)

This defect gets a CLI-side guard **in v0.3.0**, before the #192 freeze. It is the only fix admitted that restarts the freeze, so everything below ships as **one PR**:

1. **CLI guard**: when a line selection (`-l`, `--include-matching`, or `--exclude-matching`) covers exactly one side of a one-for-one replacement pair, fail with a non-zero exit naming the pair and the remedy (match text both lines share, or address both lines). Echo the existing grouped-replacement guard's message style (`cannot partially select lines N-M: ...`). All three mechanisms share the exemption at `_lines.py:179`, so one check there covers them all.
[...]

PR #226, opened 2026-08-09
feat: add three hard eval tasks where bare Git has no per-hunk answer

wkentaro:
## Summary

- Add three version 2 eval tasks targeting the situations where non-interactive bare Git has no per-hunk selection at all, so the paired comparison measures what git-hunk adds rather than what both tools share:
  - `split_single_hunk` — two intents inside one hunk, closer than the diff context width, each committed on its own.
  - `separate_formatter_noise` — a bug fix sharing a hunk with formatter churn; the fix and the reformat become separate commits.
  - `pick_duplicate_hunk` — a Duplicate Hunk group whose members are byte-identical; one ships, its twin stays in the worktree. Changed-line sets cannot tell the twins apart, which is exactly the collision the complete `HEAD` snapshot exists to catch.
- Each task has a deterministic golden solver and adversarial boundary proofs, and ADR 0004 records the three as version 2 tasks. Two adversarials replay the exact one-sided-match regression from the first eval run.
- Extract a shared `find_hunk` helper into `eval/harness.py`; the pre-existing `separate_mixed_hunks` task adopts it in place of its local copy, and each new fixture guards its premise (single hunk, shared churn, `conditional` twins) so a task fails loudly instead of silently degrading if diff topology drifts.
- Fix the skill and docs guidance on partial selection: unless a lone half is the goal, a partial selection of a one-for-one replacement must cover both its `-` and `+` line — with `-l`, `--include-matching`, and `--exclude-matching` alike — and the pattern must not also hit other changed lines in the hunk. The README and domain glossary now state the same one-sided behavior. The first eval run caught the agent matching `session.get(url)`, which hits only the deletion side, so git-hunk silently committed a deletion-only half. Issue #225 tracks whether the CLI should guard this itself.

## Measured effect

`claude-sonnet-5`, one paired run per task.

| Task                     | git-hunk               | bare-git                 |
| ------------------------ | ---------------------- | ------------------------ |
| split_single_hunk        | PASS · 3c · 4t · $0.05 | PASS · 16c · 17t · $0.13 |
| separate_formatter_noise | PASS · 3c · 4t · $0.06 | PASS · 19c · 20t · $0.21 |
| pick_duplicate_hunk      | PASS · 3c · 4t · $0.04 | PASS · 10c · 11t · $0.09 |

The git-hunk variant stays at the 3-call floor on all three tasks. The bare-Git agent passes only by brute force: stash round-trips, hand-built `git apply --cached` patches, and full-file rewrites, at 3–6× the tool calls. Single paired runs, so treat deltas as directional; the cost column carries the usual prompt-cache ordering caveat. The runs predate the review-pass polish on this branch; the fixtures' diff topology is unchanged, so the numbers still describe the shipped tasks.

Before the skill fix, `separate_formatter_noise` failed on the git-hunk variant:

| Task                     | git-hunk                         | bare-git                 |
| ------------------------ | -------------------------------- | ------------------------ |
| separate_formatter_noise | FAIL partition · 3c · 4t · $0.07 | PASS · 18c · 19t · $0.14 |

The failure reproduces against the CLI directly: `--include-matching 'session.get(url)'` selects only the `-` line (the `+` line reads `session.get(url, timeout=30)`), and git-hunk commits the deletion-only half without warning. With the amended skill the agent switched to `-l 4-5`, covering the pair.

## Test plan

- [x] `make test` (731 passed) and `make lint` pass in the worktree
- [x] New goldens pass and all eight new adversarials fail at their declared grader boundary in `tests/eval/deterministic_test.py`
- [x] `make eval` passes the git-hunk variant on all three new tasks with the numbers above
- [x] Every rebuilt commit passes ruff and `tests/eval` in an isolated worktree

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@ and this project adheres to
 
 ### Added
 
+- Three hard agent eval tasks covering the cases where non-interactive bare Git
+  has no per-hunk answer: splitting two intents inside one hunk, lifting a fix
+  out of formatter churn that shares its hunk, and committing one member of a
+  Duplicate Hunk group while its identical twin stays in the worktree (#226).
 - Agent evals end with a Markdown summary table comparing git-hunk and bare Git
   per task, with outcome, turns, and cost per cell, a legend for the failure
   reasons that occurred, and the prompt-cache caveat that makes the cost column
@@ -48,6 +52,11 @@ and this project adheres to
 
 ### Changed
 
+- The core skill, README, and domain glossary warn that a partial selection of
+  a one-for-one replacement covers only the lines it names, after an eval run
+  showed a one-sided match silently committing a deletion-only half. The
+  warning is documentation; a guard in the CLI itself is tracked by #225
+  (#226).
 - Ask for a one-line-per-commit close in the core skill, so the agent's final
   report stops restating the diff it just committed (#220).
 - Condense the core skill to the rules an agent acts on, trimming the text it
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -32,7 +32,8 @@
 
 - **Change group**: a maximal run of changed `-` and `+` lines with no context line
   between them. Partial line selection is unrestricted for pure additions, pure
-  deletions, and one-for-one replacements. A larger grouped replacement is selected
+  deletions, and one-for-one replacements; a one-sided selection of a one-for-one
+  replacement applies only that half. A larger grouped replacement is selected
   as a whole or not selected.
 
 - **Whole-file hunk**: a tracked Hunk with no `@@` text range: a binary change, a mode-only
--- a/README.md
+++ b/README.md
@@ -168,11 +168,13 @@ expressions). Both are repeatable and OR'd, case-sensitive, and error if nothing
 matches. They are mutually exclusive with `-l` and with each other.
 
 Line selection accepts any subset of a pure addition, pure deletion, or
-one-for-one replacement. A grouped replacement with multiple deleted or added
-lines must be selected as a whole or not selected. Numeric range endpoints are
-checked against the Hunk before expansion, and no-newline state is preserved for
-each patch side. Submodule pointer changes and whole-file Hunks do not support
-line selection. Select the Hunk as a whole.
+one-for-one replacement. Selecting one side of a one-for-one replacement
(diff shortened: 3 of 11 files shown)


wkentaro:
## Final verdict (Fable reviewer)

**VERDICT: recommend-merge** · **Confidence: high**

Independent final pass after five review-and-fix rounds (six parallel reviewers per round, ~19 findings fixed). The reviewer re-read the full diff, the eval framework, and the relevant git-hunk internals, and re-ran everything: `make test` (731 passed), `make lint` (clean), the 11 new golden/adversarial deterministic tests in isolation, and the first rebuilt commit alone in a fresh worktree.

**Rationale.** All three tasks are correctly constructed and their declared adversarial boundaries match the grader's actual first-failing check: `--include-matching` is per-line substring matching with no pairing requirement (`git_hunk/_lines.py`), so both one-sided-match adversarials genuinely produce a half-replacement and fail `partition`; and because `CommitSpec.changes` is a frozenset, the two `pick_duplicate_hunk` wrong-twin adversarials pass the changed-line check and correctly fail at `final-tree` — a well-chosen proof of why the HEAD-snapshot grader exists. Each fixture guards its own premise (single hunk via tuple unpack, shared-hunk assertion, `conditional` id_stability check), so topology drift fails loudly. The docs/skill changes accurately describe the CLI's real one-sided behavior, the reproducing regression is captured as an adversarial, and issue #225 is open and matches the deferral. The three commits are clean, well-scoped, and individually green.

**Residual risks for the human merger:**
- The measured-effect table is one paired run per task and predates the review-round polish; the PR body discloses both caveats — the numbers are directional only.
- The one-sided-selection hazard remains reachable at the CLI (docs-only mitigation here); merging accepts that #225 stays open as the real guard.
- `pick_duplicate_hunk`'s twin disambiguation depends on Git's default funcname heading filling `context_before`; the fixture guard raises loudly if that drifts, but a future Git default-driver change would surface as an eval fixture error rather than a silent miss.
- `find_hunk` matches its needle against the full `show` rendering including context lines; the helper's comment warns callers, and current needles are unambiguous, but future task authors must keep needles unique to changed lines.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #237, opened 2026-08-09
feat: sample each eval task variant N times and report spread

### LINKED ISSUE #223: Eval: run each task variant repeatedly and report spread instead of one sample

## Parent

Surfaced while reviewing #220.

## What to build

An eval run should be able to sample each task variant more than once and report
the spread, so a reader can tell a real workflow change from run-to-run noise.

Today every cell in the summary table is a single sample. ADR 0004 already says
the per-variant count is "a demonstration tally and not a success rate over
repeated trials", but the turn and cost columns carry the same weakness without
saying so, and they are the columns people quote.

The gap has a concrete cost. While measuring #220, consecutive runs of an
unchanged tree produced 12 and 13 tool calls, and the difference was not the
change under test: the agent had occasionally spent one extra call probing
`which git-hunk` before loading the skills. Deciding whether that was a
regression took several extra full runs, each of which spends real model usage.
A repeat count would have answered it in one invocation.

## Acceptance criteria

- [ ] The runner accepts a repeat count and runs every selected task variant that
      many times against the same prepared initial state.
- [ ] The summary table reports, per cell, a measure of spread alongside the
      central value for the graded outcome, tool calls, turns, and cost, in a
      form that stays readable as Markdown.
- [ ] The pass column distinguishes "passed every repeat" from "passed some", so
      a flaky variant cannot read as a clean pass.
- [ ] The run manifest records every individual repeat, not only the summary, so
      a run stays diagnosable after the fact.
- [ ] A repeat count of one produces the current single-sample behaviour and
      output, so existing usage is unaffected.
- [ ] ADR 0004 is amended to describe what a repeated run reports and what it
      still does not claim.

## Blocked by

wkentaro:
Closes #223.

## Summary

- `python -m eval --repeat N` samples every selected task variant N times. The
  task is prepared once per run, so the repeats differ only in the model run.
- Each summary cell reports the median of its repeats with the observed range in
  brackets, for tool calls, turns, and cost. The range is omitted when the
  minimum and maximum render identically, so a cell widens only where a reader
  can see noise.
- The pass column reads `PASS k/k`, `MIXED j/k`, or `FAIL 0/k`. A variant that
  failed any repeat changes word, not just fraction, and the total row keeps a
  mixed task out of its pass count while stating how many tasks were mixed.
- The run manifest records every repeat as its own entry with a repeat index and
  its own `.rN` trace and transcript, next to a run-level repeat count.
- `--repeat 1` is the default and reproduces the previous output and artifact
  names byte for byte.
- ADR 0004 gains a section on what a repeated run reports and what it still does
  not claim; `CONTEXT.md` gains Repeat, Spread, and Mixed cell.

## Why these two formats

**Spread as median with the observed range.** At the repeat counts this can
afford, a standard deviation would be false precision, while the minimum and
maximum are the whole of what was observed rather than a summary of it. The
range is dropped when every repeat renders the same, which keeps a stable cell
as narrow as it is today and spends width only on the metric that actually
moved.

**Partial success as a different word.** `MIXED j/k` cannot be skimmed as a
pass the way a `PASS 2/3` could, and it names every failure reason that
occurred, in grader order, so the existing legend still glosses them.

The exit-code gate now reads every repeat, so a subject variant that passes 2 of
3 exits nonzero. That is a behavior change for anything gating on the exit code
and is documented in the ADR.

## Rendered output

`--repeat 1` (unchanged from `main`):

```
| Task                      | git-hunk                    | bare-git                                 |
| ------------------------- | --------------------------- | ---------------------------------------- |
| split_refactor_vs_feature | PASS · 11c · 12t · $0.21    | FAIL order · 8c · 9t · $0.18             |
| separate_mixed_hunks      | PASS · 9c · 10t · $0.19     | FAIL leftover-worktree · 7c · 8t · $0.15 |
| **total**                 | **2/2 · 20c · 22t · $0.40** | **0/2 · 15c · 17t · $0.33**              |

bare-git runs second and may read cache written by the git-hunk run; costs are not order-neutral.

- order: a required commit order constraint is violated
- leftover-worktree: tracked worktree files do not match the expected state
```

`--repeat 3`:

```
| Task                      | git-hunk                                                   | bare-git                                                                     |
| ------------------------- | ---------------------------------------------------------- | ---------------------------------------------------------------------------- |
| split_refactor_vs_feature | PASS 3/3 · 12c [12-13] · 13t [13-14] · $0.22 [$0.21-$0.23] | MIXED 1/3 partition, order · 19c [18-21] · 20t [19-22] · $0.33 [$0.31-$0.36] |
| separate_mixed_hunks      | PASS 3/3 · 9c · 10t · $0.19                                | FAIL 0/3 leftover-worktree · 15c [14-16] · 16t [15-17] · $0.28 [$0.26-$0.29] |
| **total**                 | **2/2 · 21c [21-22] · 23t [23-24] · $0.41 [$0.40-$0.42]**  | **0/2 (1 mixed) · 34c [32-37] · 36t [34-39] · $0.61 [$0.57-$0.65]**          |

bare-git runs second and may read cache written by the git-hunk run; costs are not order-neutral. Only the first repeat starts cold, so a cost range mixes cache warmup with run-to-run noise.

- partition: commits do not match the required change groups
- order: a required commit order constraint is violated
- leftover-worktree: tracked worktree files do not match the expected state
```

## Not exercised against a live model

**This change has not been run against a live eval.** `make eval` spends real
model usage per run and a repeated run spends N times as much, so verification
is entirely through the deterministic, model-free tests in `tests/eval/` — the
separation ADR 0004 mandates. The rendered tables above come from
`render_summary` itself and are pinned in tests rather than hand-computed.
Someone should run `make eval EVAL_ARGS="--repeat 2 --task drop_debug_lines"`
once before relying on the numbers.

## Test plan

- [x] tests added: `tests/eval/summary_test.py` (median and range, range omitted
      when repeats render alike, `MIXED` never reads as a pass, reasons in
      grader order, even-count median, sub-cent floor inside a range, partial
      usage reporting, mixed task excluded from the total)
- [x] tests added: `tests/eval/runner_test.py` (`--repeat` in help, rejects 0,
      one prepared task feeds every repeat, `.rN` artifacts, per-repeat manifest
      entries, the gate fails on a mixed subject variant)
- [x] `--repeat 1` parity is pinned by the pre-existing table tests, unchanged
- [x] `make test` (733 passed) and `make lint`
- [x] each commit verified independently in a temporary worktree
- [x] `python -m eval --help` and `--repeat 0` smoked; no live eval run

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,11 @@ and this project adheres to
 
 ### Added
 
+- `python -m eval --repeat N` samples every selected task variant N times from
+  the same prepared state. Each cell reports the median with its observed range
+  for tool calls, turns, and cost, a variant that failed only some repeats reads
+  `MIXED j/k` instead of a pass, and the run manifest keeps every individual
+  repeat. `--repeat 1` stays the default and is unchanged (#237).
 - Three hard agent eval tasks covering the cases where non-interactive bare Git
   has no per-hunk answer: splitting two intents inside one hunk, lifting a fix
   out of formatter churn that shares its hunk, and committing one member of a
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -79,10 +79,21 @@
   code. The git-hunk variant is the subject; bare Git is the comparison baseline, so its
   graded outcome is evidence and never fails the run.
 
-- **Agent demonstration**: one side-by-side scenario run by the same pinned agent
+- **Agent demonstration**: a side-by-side scenario run by the same pinned agent
   from the same Repository state with bare Git and with the git-hunk toolchain. It
   illustrates the difference between the resulting commit series but is not a
   statistical benchmark.
+
+- **Repeat**: one complete sample of a task variant. A run can take several from
+  the same prepared Repository state; they differ only in the model run.
+
+- **Spread**: what a summary cell reports across its Repeats, as the median with
+  the observed minimum-to-maximum range. It describes run-to-run variation, not
+  a confidence interval.
+
+- **Mixed cell**: a summary cell whose task variant passed some Repeats and
+  failed others. It is neither a pass nor a plain failure and reads as
+  `MIXED j/k`.
 
 ## Key decisions
 
--- a/docs/adr/0004-agent-eval-design.md
+++ b/docs/adr/0004-agent-eval-design.md
@@ -36,7 +36,8 @@ distribution metadata.
 
 ### Compare git-hunk with bare Git
 
-Every selected task runs twice, first with `git-hunk` and then with bare Git.
+Every selected task runs both variants, first with `git-hunk` and then with bare
+Git, once per repeat.
 Both variants use the same initial Repository state, model, task-specific
 prompt, and grader. The task is built once, then its complete repository is
 copied back to the same temporary checkout path before each variant so Git
@@ -166,13 +167,14 @@ not `passed`, because it reports the exit-code gate below rather than every
 graded outcome.
 
 The run ends with one Markdown table: a row per task, a column per variant, and
-a cell holding that variant's outcome, turn count, and cost. A multi-task run
-adds a total row whose per-variant count is how many of those distinct tasks
-that variant got right. Because each cell is one sample, that count is a
-demonstration tally and not a success rate over repeated trials. Grading is the
-headline, so a failed cell names the grader's failure reason verbatim and a
-legend below the table glosses only the reasons that occurred. That keeps one
-vocabulary across grader, manifest, transcript, and table. The legend is a
+a cell holding that variant's outcome, tool-call count, turn count, and cost. A
+multi-task run adds a total row whose per-variant count is how many of those
+distinct tasks that variant got right. That count is a demonstration tally and
+not a success rate over repeated trials, whether the run sampled each cell once
+or the several times the next section allows. Grading is the headline, so a
+failed cell names the grader's failure reason verbatim and a legend below the
+table glosses only the reasons that occurred. That keeps one vocabulary across
+grader, manifest, transcript, and table. The legend is a
 Markdown list: consecutive bare lines collapse into one rendered paragraph,
 which would make a pasted legend unreadable. A cell whose run reported no usage
 shows the outcome with its metrics collapsed, and the total then states how many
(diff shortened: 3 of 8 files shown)


wkentaro:

An independent Fable 5 agent reviewed this branch after the implementing agent's own `/review-fix` loop, with instructions to judge fitness to merge rather than re-litigate nits, and to verify the PR's load-bearing claims rather than accept them.

> - **Full deterministic suite**: 733 passed, 0 failed, run with the worktree venv. One initial failure (`tests/e2e/commit_test.py::test_commit_reports_clean_error_when_staged_diff_fails`) turned out to be my shell exporting `FORCE_COLOR=3`, which makes `git-hunk` colorize `error:` in captured stderr; with that variable scrubbed the suite is fully green. Not a branch defect.
> - **Lint gates**: `ruff format --check`, `ruff check`, `ty check`, `mdformat --check`, `typos` all clean.
> - **`--repeat 1` byte-identity**: I loaded main's `eval/summary.py` and the branch's side by side and rendered a 650-combination grid of single-sample shapes (pass/fail, every failure reason, missing usage, sub-cent cost, one and two tasks). Every rendering was byte-identical. The `.rN` artifact suffix is applied only when `repeats > 1`.
> - **One prepared state per task**: `prepare_task` wraps the repeat loop, and `PreparedTask.run_and_grade` re-copies the snapshot into the checkout before every run, so repeats share the prepared state and each run still starts clean.
> - **No failed repeat can render `PASS`**: `_format_outcome` selects `PASS` only via `_cell_passed` (all repeats passed). I also rendered N=1 and N=3 tables by eye: ranges appear only where values differ, `MIXED 1/3 order` and `FAIL 0/3 leftover-worktree, solver-error` both gloss correctly in the legend, and the total row's `(1 mixed)` note reads naturally.
> - **ADR consistency**: the amendment keeps the "demonstration tally, not a success rate" framing, the one-vocabulary rule, the `<$0.01` floor inside ranges, and extends the cache caveat correctly for repeats.

On the four judgment calls raised in the PR:

1. **The stricter exit gate is not a breaking change.** `--repeat` is new and defaults to 1, and at N=1 the gate is exactly the old gate. No existing CI invocation changes behavior; only someone opting into repeats gets the all-repeats gate.
2. **The `/review-fix` oscillation exit left nothing real unfixed** — the sixth reviewer's re-proposals of deliberately dropped edits are the oscillation the exit condition exists for.
3. **The changelog entry is right for consistency** — the same `[Unreleased]` section already documents the eval summary table, so omitting this one would be the anomaly.
4. **`MIXED 2/3 solver-error` is a reasonable decline.** A distinct `ERROR` word would break the ADR's one-vocabulary rule, and the gate already fails the whole run on any `solver-error`, so the label cannot launder an infrastructure failure into a green run.

No defect was found, so there is no reproduction to report.

### Outstanding, all non-blocking

- The changelog entry does not mention that a repeated run exits nonzero unless every subject repeat passes. The ADR and PR body cover it; one clause ("a mixed subject run exits nonzero") would put it where CI users look first.
- The ADR sentence claiming `--repeat 1` reproduces the prior single-sample run "output and artifact names included" is slightly overbroad: `run.json` now carries a run-level `repeats` field and a per-task `repeat` field even at N=1. Printed output and artifact names are byte-identical; the manifest is additive-only.
- `_format_spread` suppresses the range by comparing rendered min/max strings, which is correct only because both renderers are monotonic in the value. That invariant is implicit and would benefit from a one-line comment.

### Reviewer's caveat, from the orchestrator

The measured verification here is deterministic only. **No live model eval was run** — repeats multiply a real per-run cost — so the repeat machinery has not been exercised end to end against `claude-sonnet-5`. The first live repeated run is the remaining unknown.

PR #238, opened 2026-08-09
test: pin duplicate group IDs across a non-member commit

### LINKED ISSUE #222: Test: pin that an unrelated commit leaves a Duplicate Hunk group's Conditional Hunk IDs unchanged

## Parent

Surfaced while reviewing #220.

## What to build

A Duplicate Hunk group's Conditional Hunk IDs must survive a `commit` of a
complete hunk that is not a member of that group. The bundled `core` skill now
tells an agent it may chain complete-hunk ID operations *before* an operation on
a Conditional Hunk ID in the same Bash call, and that permission rests entirely
on this invariant. Nothing in the suite currently fails if it breaks.

The invariant holds today. Committing a complete hunk in one file leaves the
Conditional Hunk IDs of a Duplicate Hunk group in another file byte-identical,
because a member's ordinal comes from its position among the group's members and
a non-member commit does not change that membership or ordering.

The existing duplicate-identity coverage only moves group *members* between
staged and unstaged state. No case commits a non-member and then asserts the
group is untouched, so a regression in how ordinals are assigned would ship
silently and would make the skill's chaining rule unsafe.

## Acceptance criteria

- [ ] A test builds a working tree holding both a Duplicate Hunk group and at
      least one unrelated complete hunk, records the group's Conditional Hunk
      IDs, commits the unrelated hunk by its Hunk ID, and asserts the group's
      IDs are unchanged.
- [ ] The test asserts on the full Hunk IDs, not on human-output prefixes, so a
      prefix-length change cannot mask a renumbering.
- [ ] The test fails if ordinal assignment starts depending on hunks outside the
      Duplicate Hunk group.
- [ ] It lives beside the existing duplicate-identity coverage and follows the
      suite's `_scrubbed_git_env` conventions.

## Blocked by

wkentaro:
An agent may queue complete-hunk ID operations ahead of an operation on a
Conditional Hunk ID in one Bash call only while committing a Hunk outside a
Duplicate Hunk group leaves that group's Conditional Hunk IDs byte-identical.
That held, but the existing duplicate coverage only moves group members between
staged and unstaged state, so nothing failed if ordinals started counting hunks
outside the group. This records the invariant in `CONTEXT.md` and pins it.

The test puts non-members in an earlier-sorting file, a later-sorting file, and
between the group's two members, commits them by Hunk ID, and asserts the
group's IDs are unchanged. IDs come from `--json`, so they are canonical and no
prefix length can hide a renumbering, and they are compared as an ordered list
so a swap between the two members counts as a renumbering too.

## Test plan

- [x] tests added: `tests/e2e/hunk_identity/duplicate_test.py::test_duplicate_group_keeps_ids_when_a_non_member_is_committed`
- [x] `make test` (721 passed) and `make lint`
- [x] mutation-checked: the new test is the only failure under six perturbations
      of the ordinal assignment in `git_hunk/_hunk.py` (offset by total hunk
      count, rank among all hunks, rank among same-file hunks, salt by pre-image
      start line, salt by earlier-file count, salt by later-file count), and the
      suite is green again once each is reverted

Closes #222

### DIFF
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -111,4 +111,6 @@
 - Moving a complete Hunk between staged and unstaged state does not change its Hunk ID.
 - A partial-line operation creates new Hunks with new Hunk IDs.
 - A Conditional Hunk ID can change when its Duplicate Hunk group changes.
+- A Duplicate Hunk group keeps its Conditional Hunk IDs when complete Hunks outside
+  the group are committed.
 - Each Hunk ID in an inventory addresses exactly one Hunk.
--- a/tests/e2e/hunk_identity/duplicate_test.py
+++ b/tests/e2e/hunk_identity/duplicate_test.py
@@ -1,3 +1,5 @@
+from pathlib import Path
+
 import pytest
 
 from ..conftest import GitHunkCLI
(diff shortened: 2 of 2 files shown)


wkentaro:

An independent Fable 5 agent reviewed this branch with one instruction that matters for a test-only PR: do not trust the claimed mutation coverage, devise mutations nobody else tried, and find a bug this test *should* catch but does not.

> I independently mutation-tested the new test against a full 721-test baseline run. Four mutations to `git_hunk/_hunk.py`, each run against the entire suite and each fully reverted afterward (backup + hash-verified restore): **(A)** salting the ordinal with the total inventory size — re-confirms the previously claimed mutation; **(B)** my own: assigning the ordinal from the member's rank among *all hunks in its file* rather than among group members — exactly the bug class issue #222 describes; **(C)** my own: ordinal from the member's rank across the *whole inventory* sorted by `(file, position)` — cross-file dependence; **(D)** my own: reversing the member sort whenever the total hunk count is odd — a sort-order perturbation that exercises the "list, not a set" swap assertion (5 hunks before the commit, 2 after, so member ordinals swap).
>
> All four turned `test_duplicate_group_keeps_ids_when_a_non_member_is_committed` red as the **sole failure out of 721** — no other test in the suite catches any of them, so this test is uniquely load-bearing for the invariant. The scenario design earns that: non-members sit in an earlier file, a later file, and physically between the two group members, and the assertions use canonical `--json` IDs in an order-sensitive list.
>
> I found no mutation in the ordinal/sort machinery that *should* be caught by this invariant but is not: perturbations that leave member relative order and ordinal values intact (position-as-salt, constant sort-key shifts) provably do not change ID bytes, so they are not violations of this invariant, and status-dependence is already pinned by the neighboring tests.

It confirmed both findings the implementing agent declined, and confirmed the worktree was restored — `shasum` of `git_hunk/_hunk.py` matches the pre-review backup and `git status --porcelain` is empty.

The absence of a `CHANGELOG.md` entry matches precedent (prior test-only merges `5561788` and `a25746d` shipped without one), and the added `CONTEXT.md` invariant line is accurate and correctly worded in the domain vocabulary even though it exceeds the letter of the acceptance criteria.

### Outstanding, neither blocking

- **Docs follow-up.** `git_hunk/skills/core/SKILL.md:35` scopes the commit guarantee to non-`conditional` IDs, which under-sells the invariant this test now pins and leaves the chaining-before-a-conditional-op permission implicit at lines 64-68. Either extend that sentence ("…and leaves the Conditional Hunk IDs of any group it did not touch valid") or add one near 64-68. **Issue #222's body should not be treated as evidence the skill already says this — it does not.** That claim in the ticket was wrong; the test is right regardless.
- **Pre-existing test gap.** The staged branch of `_compute_worktree_position` has an untested cross-file guard: deleting `peer.file != hunk.file` leaves the whole suite green. Covering it needs a staged Duplicate Hunk group member plus an unstaged hunk in another file, reachable via `git-hunk stage` but not `commit` (which refuses a dirty index). Correctly deferred — out of scope for #222.

Issue #58, opened 2026-06-18
Optional per-commit validation hook (--verify) for staged snapshot


wkentaro:
> *This was generated by AI.*

## Problem

git-hunk's whole purpose is producing a clean series of focused commits where each commit is independently valid. But nothing enforces that invariant — an agent can stage a partial hunk that leaves the file unparseable (e.g. a selection that drops a closing brace) and only find out later.

## Proposal

Add an optional validation command that runs against the staged-tree snapshot and refuses the stage (non-zero exit) on failure:

```bash
git-hunk stage 66279ce -l ^115-119 --verify 'python -m py_compile path/to/file.py'
```

Behavior:

- Materialize the would-be-staged content (the same snapshot `preview` would emit), run the `--verify` command against it, and only commit the stage to the index if the command exits 0.
- On failure, leave the index untouched and report the command's output.

This bakes the "each commit builds" discipline into the tool instead of relying on the agent to remember to check.

## Notes

- Depends conceptually on materializing the staged file state (see the `preview` issue); could share that machinery.
- Could later be promoted to a configurable default (e.g. a repo-level verify command) but a per-invocation flag is the minimal version.
- One of a set of agent-usability ideas from the same session.

wkentaro:
> *This was generated by AI during triage.*

## Triage Notes — deferred, blocked by #57

This is a sound request, but it is **not yet ready for an agent** because it is built directly on machinery that does not exist yet.

**What's established:**
- `--verify <cmd>` should materialize the would-be-staged snapshot, run `<cmd>` against it, and commit the stage to the real index **only if `<cmd>` exits 0**; on failure, leave the index untouched and report output.
- This is the same "materialize staged content without touching the index" primitive that **#57 (`preview`)** introduces. #58 should reuse it, not reinvent it.

**Why deferred:** briefing #58 now would have to guess the shape of a primitive #57 is about to define. Sequence it **after #57 lands**, then reuse `#57`'s `materialize_staged_content`.

**Design questions to settle once #57 exists (the genuinely unresolved part):**
- **Where does `<cmd>` run?** Against a temp worktree populated with the would-be-staged content of the affected file(s), so a relative path like `python -m py_compile path/to/file.py` resolves? Or is the snapshot handed to the command another way (stdin, a temp file path in an env var)? This is the core UX decision and affects how multi-file selections behave.
[...]

Issue #57, opened 2026-06-18
preview: emit resulting staged file content without touching the index


wkentaro:
> *This was generated by AI.*

## Problem

An agent's commit discipline is "each commit is independently valid" — the build/tests would pass at every commit. After a partial stage, the only way to check that the *resulting* file is valid (parses, compiles, passes tests) is to shell out:

```bash
git show :path/to/file.py | python -c 'import sys, ast; ast.parse(sys.stdin.read())'
```

That works but is a manual detour, and `git show :file` has its own pathspec quirks (e.g. it wants the index path, not a relative path). There is no first-class way to see what a file *would* look like after a given selection before committing.

## Proposal

Add a preview that emits the resulting file content for a hypothetical stage selection, without mutating the index:

```bash
git-hunk preview 66279ce -l ^115-119          # full resulting file content
git-hunk preview 66279ce --exclude-matching 'c1205_20260605'
```

This lets the agent validate the post-stage state (syntax check, run a linter, even run tests) *before* `stage` + `commit`, instead of stage → inspect → maybe-unstage. It closes the goal-driven loop: select → verify → commit.

## Notes

- Distinct from showing the staged *diff*; the ask is the resulting file *content* so it can be fed to a validator.
- Pairs naturally with a per-commit validation hook (separate issue) but is useful on its own.
- One of a set of agent-usability ideas from the same session.

wkentaro:
> *This was generated by AI during triage.*

## Agent Brief

**Category:** feature
**Summary:** Add a `preview` command that emits the resulting file content for a hypothetical stage selection **without mutating the index**, so an agent can validate the post-stage file (parse/lint/test) before committing.

**Current behavior:**
To check what a file would look like after a partial stage, the agent must stage, then `git show :path | validator`, then maybe unstage. `git show :file` also has pathspec quirks (wants the index path). There is no first-class "what would this selection produce" view.

**Desired behavior:**
`git-hunk preview <id> [selection flags]` writes the **full resulting file content** (what the file would be in the index after applying that selection) to stdout, leaving the real index and working tree untouched:

```
git-hunk preview 66279ce -l ^115-119
git-hunk preview 66279ce --exclude-matching 'c1205_20260605'
```

- Accepts the **same selection flags as `stage`** (`-l`, and #55's `--include/exclude-matching` once landed). With no selection flag, it previews staging the whole hunk.
[...]

PR #239, opened 2026-08-09
feat: materialize would-be-staged content without touching the index (groundwork for #58)

wkentaro:
## Summary

Adds one internal seam and the plumbing it needs. Nothing calls it yet.

`materialize_staged_content(hunks, diff_output, worktree_root=...)` returns the
exact bytes one file would hold in the index after a selection was staged. It
applies the selected patch to a throwaway copy of the index via `GIT_INDEX_FILE`
and reads the blob back out of that copy, so the real index and working tree are
untouched. (The blob does land in the object database, unreferenced until
`git gc` reclaims it; the docstring says so rather than claiming "changes
nothing".)

The seam carries its own guards, so a caller gets a diagnosis instead of an
opaque `git apply` failure: the selection must cover exactly one file, must not
be binary, a type change, a submodule, or mode-only, and must not remove the file
from the index. It returns raw `bytes`, so non-UTF-8 content, CRLF, and a missing
trailing newline survive unchanged.

This is groundwork for `stage --verify` (#58): automatic verification at stage
time, where the check is not something the agent has to remember.

## Why this is not the `preview` command #57 asked for

#57 asked for a user-facing `git-hunk preview` command. An earlier revision of
this PR shipped it. A controlled ablation says it does not earn its place, so
this PR keeps only the machinery.

`claude-sonnet-5`, 30 samples, including 10 + 10 repeats on a scaled trap fixture
where the tempting exclusions provably stage unparseable Python:

| Variant           | Outcome | Tool calls (median) | Cost (median) | Broken commits |
| ----------------- | ------- | ------------------- | ------------- | -------------- |
| with `preview`    | 10/10   | 8                   | $0.09         | 0              |
| without `preview` | 10/10   | 5                   | $0.07         | 0              |

Zero outcome difference. The command cost +3 tool calls and +$0.02 median for
verification the model did not need: the agents without it avoided the trap by
selecting carefully in the first place, not by checking afterwards.

So the user-facing surface is declined, and the seam that a real fix can build on
is kept. The real-world incident evidence supports verification happening
automatically at stage time (#58), not an extra command an agent must think to
run.

## Commits

1. `refactor(git)` — add a bytes-returning core under `run_git` and thread an env
   overlay through, layered onto `os.environ` rather than replacing it, so git
   keeps `PATH`, `HOME`, and the caller's config. Behavior is unchanged.
2. `feat` — `_staged_content.py`, plus `scratch_index` and `read_index_blob` in
   `_git.py`.

Each commit builds and passes the full suite on its own.

## Notes

- The module is `git_hunk/_staged_content.py`, not `_preview.py`, and the error
  type is `StagedContentError`. With no `preview` command, "preview" in the error
  strings would name something that does not exist; these messages surface
  through `stage --verify` later.
- `docs/adr/0002` gains the `git ls-files` that reads the materialized blob to
  the list of calls that pass `--literal-pathspecs`, and corrects a pre-existing
  omission (`git rm`).
- No `CHANGELOG.md` entry: there is no user-facing change to document.

## Test plan

- [x] tests added: `tests/unit/_staged_content_test.py` (17 cases) — byte
      exactness for whole-hunk, include, and exclude selections; non-UTF-8
      content with no trailing newline; CRLF; equality with what a real staging
      apply produces across three selection forms; building on an already-staged
      index; unborn HEAD; the real index and worktree unchanged; and every
      rejection path (binary, mode-only, empty tracked addition, two files,
      index deletion)
- [x] `make test` (763 passed) and `make lint` green on each of the two commits

Refs #57

### DIFF
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -72,6 +72,12 @@
   `{"text": s}` when the bytes are valid UTF-8, else `{"bytes": base64}`. Always an object
   (even for valid UTF-8) so consumers have one code path. The ripgrep `--json` idiom.
 
+- **Materialized staged content**: the exact bytes one file would hold in the index
+  after a given selection was staged. Produced by applying the selected patch to a
+  throwaway copy of the index, so the real index and worktree never change. Internal
+  only (`_staged_content.materialize_staged_content`): it is the unit a validator
+  would check before a partial selection is committed, not something a command emits.
+
 - **git-hunk toolchain**: the git-hunk CLI together with the bundled `core` and
   `logical-commits` skills as presented to an agent.
 
--- a/docs/adr/0002-repository-path.md
+++ b/docs/adr/0002-repository-path.md
@@ -64,7 +64,8 @@ An operand names one exact changed Repository path. Directories, globs, and Git
 pathspec magic are not expanded. A leading `./` and internal `..` components are
 normalized; absolute paths and paths that escape the worktree are rejected.
 
-`git add` and `git restore` are the only calls that hand Git a path as a pathspec, so
+`git add`, `git rm`, `git restore`, and the `git ls-files` that reads a materialized
+blob out of the scratch index are the only calls that hand Git a path as a pathspec, so
 they alone pass `--literal-pathspecs`. The path they receive is `Hunk.file`, taken from
 Git's own diff output rather than from the operand, so this guards any filename
 containing pathspec punctuation however the Hunk was selected. The flag is deliberately
--- a/git_hunk/_git.py
+++ b/git_hunk/_git.py
@@ -1,4 +1,6 @@
+import os
 import subprocess
+from collections.abc import Mapping
 from dataclasses import dataclass
 from typing import Final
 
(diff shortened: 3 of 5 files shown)


wkentaro:
## Independent review verdict: MERGE

An independent reviewer (Fable, no part in authoring this change) reviewed the PR empirically in a throwaway repo. No correctness, cross-platform, or index-safety defects found; full suite passes (747 tests).

**Acceptance criteria from #57 — all confirmed empirically:**
- Partial selections (`-l` include/exclude, `--include/--exclude-matching`) print the staged-selection content with the real index and worktree byte-identical afterward (verified via `git diff`/`git diff --cached` checksums before/after each call).
- Output is validator-ready byte-for-byte: non-UTF-8 bytes, missing trailing newline, and CRLF content all round-trip exactly; `preview | python -c 'ast.parse(...)'` works. `click.echo(bytes, nl=False)` writes through the binary stream, so no CRLF mangling on Windows.
- Whole-hunk preview is `cmp`-equal to `git show :file` after a real stage, including multi-hunk and preview-on-top-of-already-staged-index cases.
- Binary, mode-only, multi-file, and index-deletion selections exit 1 with exactly 0 bytes on stdout and a clear message + tip on stderr.
- `materialize_staged_content` (`git_hunk/_preview.py`) carries all guards and imports nothing from `_cli` — reusable for #58 as specified.

**Scratch-index mechanics verified sound:** respects ambient `GIT_INDEX_FILE` (works inside hooks/rebase), env overlay layered onto `os.environ`, survives `core.splitIndex`, linked worktrees, and subdirectory invocation; a concurrent index change fails cleanly into `PreviewError` with no corruption path.

**Spec deviations all judged acceptable:** guards in `_preview.py` behind `PreviewError` (so #58 gets typed rejections, not opaque `git apply` stderr); mode+text files preview while mode-only errors; multiple hunk IDs within one file (strict superset, verified byte-equal to staging both); the honest "unreferenced loose objects until `git gc`" disclosure confirmed accurate.

**Non-blocking note:** bulk preview loops grow the object database until `git gc` — inherent to `git apply --cached`, documented; worth remembering when building #58.

wkentaro:
**Post-rebase re-review (independent, Fable): MERGE — verdict stands.** After two rebases onto moving main (#226, #237 merges), an independent reviewer re-verified: no conflict markers repo-wide, CHANGELOG/CONTEXT resolutions purely additive with nothing lost or duplicated, ancestry correct, 772 tests green, and the preview contract re-confirmed empirically (partial selections, untouched index/worktree, empty-stdout error cases, byte-exact non-UTF-8 output). Note: the PR body's "747 passed" is stale post-rebase; the branch is at 772.

wkentaro:
**Independent re-review of the rework (Fable): MERGE.** Verified empirically: the strip is complete (every remaining "preview" match is the pre-existing `--dry-run` wording, byte-identical to main); `_staged_content.py` is a pure internal seam (no importers in `git_hunk/`, no CLI imports, `StagedContentError` + tip is a sensible surface for #58); the ported API tests cover the full old e2e checklist (byte exactness incl. non-UTF-8/CRLF/no-trailing-newline, untouched index+worktree, all guards, already-staged base, unborn HEAD) and pass; a direct temp-repo materialization spot-check round-tripped exactly. 763 tests green, and both commits green individually. Non-blocking notes: submodule/type-change guards and the unmerged/gitlink branches lack direct tests.

wkentaro:
Closing per the ablation outcome. This PR's rework left only internal groundwork for #58, and #242 no longer depends on it (its golden solver verifies via stage → `git show :file` → `ast.parse`). With #58 also closed on the same evidence — 30 ablation samples, zero broken commits without any verification, and the one observed incident class (#226) covered by the skill warning plus the #225 guard — this seam has no consumer. The branch and its review survive in history; if the eval's `broken-commit` tripwire ever fires in practice, this is the starting point to revive.

PR #242, opened 2026-08-09
feat(eval): grade a commit that does not parse, and a task that tempts one

wkentaro:
## Summary

Stacks on #239 (base `feat/57-preview-staged-content`); GitHub retargets this to
`main` when #239 merges. Nothing here depends on #239 any more — the stack only
keeps the review order.

- New grader criterion `broken-commit`: for every commit in `base..HEAD`, every
  `.py` blob in that commit's own tree must parse. It runs **before** the
  partition check, because a commit nobody can check out and run is a worse
  fault than a wrongly grouped one, and reporting it as `partition` would name a
  line-set mismatch when the real defect is a file that does not compile. It
  reads each commit's own tree, so a broken intermediate state cannot hide
  behind a correct final tree, and the detail names the commit, the path, and
  the syntax error.
- New eval task `commit_parseable_subset`. One hunk interleaves the change worth
  keeping with debug scaffolding:

  ```text
  4 +        if row.voided:
  5 +            print("DEBUG", "voided", row)
  6 +            continue
  7 +        if total > 1000:
  8 +            print("DEBUG", "large total", total)
  ```

  Only lines 4 and 6 belong in the commit, and every tempting whole-noise
  selection stages Python that does not parse:
  `--exclude-matching 'print("DEBUG"'` and `-l ^5,^8` both strand line 7, the
  `if` header whose only body is a debug print, and any single range covering
  both prints (`-l ^5-8`) also swallows the `continue` the kept change needs.
  The trap is structural, not the one-sided replacement-pair case the `core`
  skill already warns about.
- The golden solver runs the stage-then-verify dance an agent has to perform by
  hand today: stage `-l 4,6`, `ast.parse` the bytes `git show :report.py` then
  returns, and commit from that verified index. That is the sequence #58 wants
  to fold into `stage --verify`; the grader criterion is the backstop either
  way, since it fails the run whether or not the agent thought to check.
- Adversarials: two `broken-commit` (the match trap and the range trap, proving
  the new boundary fires ahead of `partition`), one `partition` (everything in
  one commit), one `leftover-worktree` (correct commit, scaffolding left
  behind).

Every pre-existing adversarial still fails at its originally declared boundary
after the ordering change — all existing task commits hold parseable Python, so
nothing shifted, and `tests/eval/deterministic_test.py` (which auto-discovers
`SCENARIOS`) proves it rather than assuming it.

## What changed since the last review

Only the golden solver, and the prose that described it. The task fixture, the
prompt, the adversarial matrix, and the grader are unchanged in what they
enforce.

The earlier golden shelled `git-hunk preview`. That command is not shipping —
see #239, where a 30-sample ablation found zero outcome difference from having
it — so the golden now performs the same check with `stage` plus
`git show :<path>` instead.

## Measured effect

The paired `claude-sonnet-5` run reported earlier (git-hunk PASS · 5 calls ·
$0.10 vs bare-git PASS · 15 calls · $0.14) was measured with the `preview`
command available to the agent, so it no longer reproduces and has been dropped
rather than restated. Both variants passed then, so the task measured workflow
rather than capability; the bare-Git agent reached the same end state only after
overwriting the dirty working tree by hand and rebuilding it from a hand-written
`git apply` patch.

The evidence that matters for this PR is that the task and criterion catch a
selection that produces a commit which does not compile, which the deterministic
adversarials prove without spending model usage.

## Test plan

- [x] `make test` (772 passed) and `make lint` green
- [x] New golden passes and all four new adversarials fail at their declared
      boundary; the grader gains two focused unit tests (a broken *intermediate*
      commit is reported as `broken-commit` even though the commit count is
      wrong too, and non-Python files are never parsed)
- [x] Each commit is green on its own

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,12 @@ and this project adheres to
 
 ### Added
 
+- Agent eval task and grader criterion for a partial selection that would
+  commit syntactically broken code: one hunk interleaves the change to keep
+  with debug scaffolding, so every tempting whole-noise selection stages
+  Python that no longer parses, and the grader now fails any commit in the
+  graded range whose own tree holds an unparsable `.py` file, ahead of the
+  commit-partition check (#242).
 - `python -m eval --repeat N` samples every selected task variant N times from
   the same prepared state. Each cell reports the median with its observed range
   for tool calls, turns, and cost, a variant that failed only some repeats reads
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -79,6 +79,11 @@
   code. The git-hunk variant is the subject; bare Git is the comparison baseline, so its
   graded outcome is evidence and never fails the run.
 
+- **Broken commit**: an eval grader verdict — a commit in the graded range whose own
+  tree holds a `.py` file that does not parse. Checked before the commit partition,
+  because a commit nobody can check out and run is a worse fault than a wrongly grouped
+  one, and checked on every commit in the series rather than only on `HEAD`.
+
 - **Agent demonstration**: a side-by-side scenario run by the same pinned agent
   from the same Repository state with bare Git and with the git-hunk toolchain. It
   illustrates the difference between the resulting commit series but is not a
--- a/docs/adr/0004-agent-eval-design.md
+++ b/docs/adr/0004-agent-eval-design.md
@@ -80,12 +80,23 @@ The grader checks the first failed invariant in this order:
 
 | Check              | Required state                                                   | Failure reason       |
 | ------------------ | ---------------------------------------------------------------- | -------------------- |
+| Commit parses      | Every `.py` blob in each commit's own tree parses                | `broken-commit`      |
 | Partition          | Actual commits match the declared changed-line groups            | `partition`          |
 | Order              | Required before and after constraints hold                       | `order`              |
 | Final commit       | The complete `HEAD` tree matches the declared files              | `final-tree`         |
 | Index              | Its tree and exact paths equal `HEAD`                            | `leftover-index`     |
 | Tracked worktree   | Exact bytes and modes match the declared tracked state           | `leftover-worktree`  |
 | Untracked worktree | Exact paths, bytes, and modes match the declared untracked state | `leftover-untracked` |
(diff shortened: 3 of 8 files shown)


wkentaro:
**Independent final review (Fable): MERGE — after one blocking fix, now applied in 153fe39.** The reviewer found `_find_broken_commit` caught only `SyntaxError`, but `ast.parse` raises `ValueError` for source it refuses to read at all (e.g. null bytes) — verified empirically: a graded commit holding `a.py` with a null byte crashed `grade()` and would have killed a paid multi-task eval run instead of failing the task as `broken-commit`. Fixed by catching `(SyntaxError, ValueError)` with a regression test (781 tests green).

Everything else confirmed: rebase integrity on both stack branches (no markers, additive changelog/glossary resolutions, correct ancestry), all three tempting selections reproduce unparseable Python against the real CLI (stranded `if` headers / swallowed `continue`), the golden preview→ast.parse→commit workflow works end-to-end, and `broken-commit` ordering before the partition check is pinned by unit test. Non-blocking notes: per-commit tree re-parsing has no blob memoization (fine at eval scale) and an intentionally-unparseable future fixture would trip the global check (documented as a revisit trigger in ADR 0004).

**Stack verdict: safe to merge bottom-to-top (#239 → #242).**

wkentaro:
**Independent re-review of the rework (Fable): MERGE.** The golden solver's new stage → read `:report.py` → `ast.parse` → commit-from-verified-index → discard sequence was executed in a temp repo: exactly one commit, final state byte-identical to spec, and the forced parse-failure path unstages cleanly with no commit created. All adversarial boundaries re-proven (5/5), full suite 772 green, every intermediate commit of the stack green (746→772). One pre-existing environmental flake noted (harness `.git` snapshot copy racing background `git maintenance` — exists on main, worth a follow-up issue to set `maintenance.auto=false` in eval temp repos). Nit: golden catches only `SyntaxError` where the grader catches `ValueError` too — unreachable here, purely defensive.

**Stack verdict: coherent, zero user-visible residue of the removed command, safe to merge bottom-to-top.**

PR #245, opened 2026-08-09
test: pin that a staged hunk ignores unstaged peers in other files

### LINKED ISSUE #241: Test: pin that a staged hunk's worktree position ignores unstaged peers in other files

## Parent

Surfaced by the independent review of #238, and confirmed there by mutation.

## What to build

A test that fails when the worktree-position calculation for a **staged** hunk
stops restricting itself to peers in the same file.

When a hunk's status is `staged`, its worktree position is computed by walking
the other hunks and adjusting for the unstaged ones that precede it. That walk
skips any peer belonging to a different Repository path. Deleting the
same-file half of that guard leaves the **entire suite green** — verified
independently during the review of #238 against all 721 tests at the time.

The consequence, if it regressed, is that a Duplicate Hunk group containing a
staged member could have its Conditional Hunk IDs shift because of unrelated
unstaged hunks in a *different* file. That is the same class of bug #222 pinned
for the commit path, left uncovered on the staged path.

The gap is pre-existing and was correctly ruled out of scope for #222: that
scenario reaches the code through `git-hunk commit`, which aborts when the index
already holds staged changes, so a staged group member cannot coexist with it.
Reaching this branch needs `git-hunk stage` instead.

## Acceptance criteria

- [ ] A test builds a working tree with a staged member of a Duplicate Hunk
      group and at least one unstaged hunk in a **different** file positioned so
      that counting it would shift the staged member's computed position.
- [ ] The test asserts on full Hunk IDs from `--json`, not on human-output
      prefixes, so a prefix-length change cannot mask a renumbering.
- [ ] Removing the same-file condition from the staged-branch peer walk makes
      the test fail. Verify this by actually making the mutation, observing the
      failure, and reverting.
- [ ] It lives beside the existing duplicate-identity coverage and follows the
      suite's `_scrubbed_git_env` conventions.

## Blocked by

wkentaro:
Closes #241.

Ordering a Duplicate Hunk group's members adds the net line delta of the unstaged Hunks ahead of a staged member, counting only the ones with the same Repository path. Dropping that condition left the entire suite green, so a staged group member could be renumbered by an edit to an unrelated path with nothing to catch it.

## Why the scenario looks the way it does

Ordinals depend only on members' *relative* order, so a miscount is invisible unless it actually reorders them. That rules out an unrelated hunk that adds lines — it only pushes the staged member further from its twin. The peer has to **delete** more lines than separate the two members, and precede the staged one.

Two details are load-bearing rather than incidental:

- The deletion is sized from the fixture's own geometry, not a literal. An earlier revision hard-coded it against today's 37-line separation with three lines of slack; widening the shared `duplicate_hunks` fixture's separator to `range(50)` made the test pass while detecting nothing.
- One unrelated path sorts before the group's and one after, matching the sibling test for #222. With only an earlier-sorting path, a half-guard that leaks later-sorting peers survives.

## Mutation evidence

Each mutation applied to the guard in `_compute_worktree_position`, run against the full suite and reverted:

| mutation | meaning | result |
| --- | --- | --- |
| `if peer.status != "unstaged":` | guard deleted | killed |
| `... or peer.file > hunk.file:` | leaks earlier-sorting paths | killed |
| `... or peer.file < hunk.file:` | leaks later-sorting paths | killed |

The first is also killed at fixture separator `range(30)`, `range(50)`, and `range(200)`, confirming the derived sizing holds beyond the geometry it was written against.

## Known gap, out of scope

The other half of the same guard stays uncovered: mutating it to `if peer.file != hunk.file:`, so staged peers get counted too, leaves all 746 tests green. Same shape as #241, pre-existing, and not what this ticket asked for. Filed separately.

## Test plan

- [x] `make test` — 746 passed
- [x] `make lint` — clean
- [x] Mutation matrix above, with `git_hunk/_hunk.py` verified byte-identical afterwards
- [x] The commit verified in isolation in a detached worktree

No `CHANGELOG.md` entry: test-only plus an internal domain-doc line, matching `3e937eb` (#238) and prior test-only merges.

### DIFF
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -125,4 +125,6 @@
 - A Conditional Hunk ID can change when its Duplicate Hunk group changes.
 - A Duplicate Hunk group keeps its Conditional Hunk IDs when complete Hunks outside
   the group are committed.
+- Ordering a Duplicate Hunk group's members counts only the unstaged Hunks with the
+  same Repository path, so an unstaged change elsewhere cannot renumber the group.
 - Each Hunk ID in an inventory addresses exactly one Hunk.
--- a/tests/e2e/hunk_identity/duplicate_test.py
+++ b/tests/e2e/hunk_identity/duplicate_test.py
@@ -2,6 +2,8 @@ from pathlib import Path
 
 import pytest
 
+from git_hunk._hunk import parse_hunk_range
+
 from ..conftest import GitHunkCLI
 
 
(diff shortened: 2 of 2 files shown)


wkentaro:

An independent Fable 5 agent reviewed this branch after the three-round `/review-fix` loop. It was given what those rounds already found, so its effort went into mutations nobody had tried — and specifically into hunting for a mutation that *should* be caught by this invariant but is not.

## Its own mutation matrix

Each applied to `_compute_worktree_position`, run against the full 746-test suite, and reverted.

| mutation | meaning | result |
| --- | --- | --- |
| `if peer.status != "unstaged":` | guard deleted (reproducing the PR's claim) | **killed** by the new test |
| `peer.status == "unstaged" or ...` | counts staged peers, skips unstaged | **killed** (sibling test) |
| count condition forced `False` | staged position never adjusted at all | **killed** (sibling test) |
| `a_start > b_start` | counts peers after instead of before | **killed** |
| `position += a_count - b_count` | delta sign flip | **killed** |
| `a_start <= b_start` | tie-break boundary widened | **survives** |
| `a_start = peer_range.new_start` | peer position read from the post-image side | **survives** |

> The hunt asked for a mutation that violates *this* invariant yet survives. I found none beyond known #246. The two survivors both still count only same-path unstaged peers — they corrupt the same-file positional logic, which the invariant line deliberately does not claim.

## On the four open questions

**The derived sizing is geometry-independent by construction**, not merely correct for the three separations tested:

> Each unrelated file's hunk has delta `1 − (second_start − first_start + 5)`, so a leaked peer moves the staged member to exactly `first_start − 4` — strictly before its twin for any fixture where `first_start < second_start` (asserted), with a 4-line margin that also defeats stable-sort ties.

**The `new_start == second_start` pin is load-bearing, not too strict.** The sizing derivation silently depends on the staged member's position equalling the pre-staging `second_start`, which breaks if the fixture ever grows nonzero-delta duplicates. The pin turns that into a loud failure instead of a silently mis-sized scenario.

**The `parse_hunk_range` oracle bound holds, and is stronger than argued in the PR.** All three headers the test parses have `old_start == new_start`, so the most plausible parser fault — swapping old and new — is invisible to this test and cannot perturb its derived values at all. Grosser faults break sibling tests loudly, which the mutation runs above corroborate.

**The `CONTEXT.md` line is accurate** — the staged branch counts exactly `peer.status == "unstaged" and peer.file == hunk.file`, the unstaged branch counts nothing (a subset, so "only" holds), and the vocabulary matches the glossary. **No changelog entry is correct**, matching `3e937eb` (#238), the identical shape.

## Outstanding, none blocking

- Two surviving mutants are worth their own coverage, both pre-existing and both outside this ticket's scope: widening the tie-break to `a_start <= b_start`, and reading the peer's position from `new_start` instead of `old_start`, each leave all 746 tests green. They are same-file positional gaps rather than cross-file ones, most cheaply closed by direct unit tests for `_compute_worktree_position`.

## Worktree integrity

Mutations ran under a backup-and-restore exception. Verified afterwards, independently:

```console
$ git -C .wt/test/241-staged-position-cross-file status --porcelain
(empty)
$ shasum git_hunk/_hunk.py
87b0954b207416b090f27882e9b0751213462f4b  git_hunk/_hunk.py
```

Issue #232, opened 2026-08-09
Decide the v0.3 admit list and freeze point


wkentaro:
> *This was generated by AI.*

## Question

Given the readiness audit's findings, exactly which trust-critical fixes enter v0.3.0 before the #192 freeze, and which are deferred with a documented caveat? Resolves into ready-for-agent issues for admitted fixes and hands the frozen scope to #192.

wkentaro:
> *This was generated by AI.*

## Resolution

Decided with the human against the #228 audit findings, with PR #226 now merged.

**Admitted into v0.3.0** (gate: PR #226 + trust-critical fixes only):

- **PR #226** — already merged; the skill-side #225 warning is on `main`.
- **#225 CLI-side guard** — the sole freeze-restarting admit. Semantics: **error + explicit override flag** (fails closed; warn-and-proceed rejected because exit-code-only agents would still get the wrong result). Ships as **one PR** bundling the CLI guard, the core-skill text update (the merged warning becomes stale once the CLI errors), a README sentence, and the changelog entry — so the #192 freeze restarts exactly once. #225 now carries the decided scope and is `ready-for-agent`.
- **README one-sided caveat** — folded into the #225 guard PR rather than a separate task.

**Rejected**: the hex-operand ID fallback note (cosmetic; not worth a docs task in the freeze scope).

**Deferred past v0.3.0** as ordinary issues with **no release caveat** — all fail loudly or are cosmetic, below the trust bar: #200, #139 (plus its newly filed sibling #248, the untracked-file `os.lstat` crash), the `git restore` preflight gap, #115, #123.

**Freeze point**: v0.3.0 = `main` (PR #226 included) + the #225 guard PR. When that PR merges, the freeze restarts once and #192 proceeds with no further CLI or bundled-skill changes. Handed off via comment on #192.

Issue #192, opened 2026-08-06
Prepare and prove the v0.3.0 release candidate


wkentaro:
> *This was generated by AI.*

## Agent brief

Prepare the exact v0.3.0 candidate after the five implementation tickets land. This ticket proves readiness. It does not create a tag or publish anything.

## Plan

Follow [the release-candidate checklist](https://github.com/wkentaro/git-hunk/blob/45758d791edbb207dc1682df04aaf45af440ce5d/docs/research/186-v0.3.0-release-candidate-checklist.md#6-correct-all-release-notes) from changelog correction through the release freeze before tagging.

## Scope

- Correct every missing or wrong Unreleased pull request reference and add all omitted user-facing changes.
- Add release-note entries for every user-facing implementation change.
- Make README, package metadata, help, JSON examples, domain text, and both bundled skills truthful.
- Resolve [README header null omits untracked entries](https://github.com/wkentaro/git-hunk/issues/163).
- Run all focused, deterministic, full-test, full-lint, package-content, metadata, install, and Repository-state checks on the exact merged candidate SHA.
- Run one uninterrupted pinned 4/4 model eval with no retry.
- Check in the redacted run evidence and change ADR 0002 to Accepted.
- Promote the changelog to 0.3.0 with the release date and restore an empty Unreleased section.

## Acceptance

- Every evidence record names one exact candidate SHA.
- Any later CLI, eval, task, or bundled-skill change invalidates the model run and requires a new qualifier.
- Test and Lint GitHub Actions are green for the exact SHA.
- The candidate worktree is clean.
- The pull request closes issue 163.
- Stop before tag creation and request the required release permission.

wkentaro:
> *This was generated by AI.*

The v0.3.0 scope is now frozen, decided in #232:

**v0.3.0 = `main` (PR #226 included) + the #225 guard PR** (CLI error + override flag, core-skill text update, README sentence, changelog — all in one PR, scoped on #225, `ready-for-agent`).

The freeze restarts once, when the #225 guard PR merges; after that, no further CLI or bundled-skill changes enter v0.3.0. Everything else surfaced by the #228 audit (#200, #139, #248, #115, #123) is deferred past v0.3.0 with no release caveat.

wkentaro:
> *This was generated by AI during implementation.*

[...]

PR #251, opened 2026-08-09
feat: reject one-sided selection of a one-for-one replacement

### LINKED ISSUE #225: Content match hitting one side of a one-for-one replacement silently commits a broken half

wkentaro:
> *This was generated by AI.*

Closes #225. Implements the guard decided in #232 (the v0.3 admit list): this is the one freeze-restarting PR for v0.3.0, handed to #192.

## What changes

- A line selection (`-l`, `--include-matching`, `--exclude-matching`) covering exactly one side of a one-for-one replacement pair now **errors with exit 1** — `cannot select one side of lines N-M: one-for-one replacement; match text both lines share, select both lines, or pass --allow-one-sided` — instead of exiting 0 after silently staging/committing/discarding a broken half. Nothing is mutated on rejection.
- New **`--allow-one-sided`** flag on `stage`, `unstage`, `discard`, and `commit` permits the deliberate lone-half case (e.g. committing a pure deletion). Passing it without a selection mechanism is a usage error (exit 2), mirroring `--regex`.
- The wider grouped-replacement guard is **not** relaxed by the flag.
- Docs updated to match: core skill passage, README, CONTEXT.md glossary, help text, changelog (**Breaking:**, folded with the #226 entry as documented-then-enforced).
- The two adversarial eval solvers that deliberately produce one-sided halves now opt in via the flag, keeping their graded failure mode intact.

## Verification

- 801 tests pass (was 746 before the change; +55 covering all four selectors × rejection/override, non-mutation on rejection asserted against HEAD/index/worktree/status, per-group error scoping, `--dry-run`, usage errors, and the wider-guard non-relaxation).
- All lints clean: ruff format/check, ty, mdformat, taplo, yamlfix, typos.
- Independently reviewed (correctness, fail-closed paths, exclude-inversion, no-newline markers, docs truthfulness) with all findings addressed.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -66,7 +66,7 @@ and this project adheres to
   lone-half case; it requires one of the selection options and does not relax
   the wider grouped-replacement rule. An eval run first caught the silent half,
   and the core skill, README, and domain glossary documented the hazard (#226)
-  before this release made it an error (#TBD).
+  before this release made it an error (#251).
 - Ask for a one-line-per-commit close in the core skill, so the agent's final
   report stops restating the diff it just committed (#220).
 - Condense the core skill to the rules an agent acts on, trimming the text it
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -57,11 +57,16 @@ and this project adheres to
 
 ### Changed
 
-- The core skill, README, and domain glossary warn that a partial selection of
-  a one-for-one replacement covers only the lines it names, after an eval run
-  showed a one-sided match silently committing a deletion-only half. The
-  warning is documentation; a guard in the CLI itself is tracked by #225
-  (#226).
+- **Breaking:** Selecting exactly one side of a one-for-one replacement now
+  exits 1 with an error naming the pair's line numbers, instead of exiting 0
+  after silently staging, committing, or discarding the deletion-only or
+  addition-only half. `-l`, `--include-matching`, and `--exclude-matching` share
+  the guard, and nothing is mutated when it fires. `stage`, `unstage`,
+  `discard`, and `commit` take a new `--allow-one-sided` flag for the deliberate
+  lone-half case; it requires one of the selection options and does not relax
+  the wider grouped-replacement rule. An eval run first caught the silent half,
+  and the core skill, README, and domain glossary documented the hazard (#226)
+  before this release made it an error (#TBD).
 - Ask for a one-line-per-commit close in the core skill, so the agent's final
   report stops restating the diff it just committed (#220).
 - Condense the core skill to the rules an agent acts on, trimming the text it
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -31,10 +31,10 @@
   Hunk group. It can change when the group changes.
 
 - **Change group**: a maximal run of changed `-` and `+` lines with no context line
-  between them. Partial line selection is unrestricted for pure additions, pure
-  deletions, and one-for-one replacements; a one-sided selection of a one-for-one
-  replacement applies only that half. A larger grouped replacement is selected
-  as a whole or not selected.
+  between them. Partial line selection is unrestricted for pure additions and pure
+  deletions. A one-for-one replacement is selected whole, or one-sided only with
+  `--allow-one-sided`, which then applies just that half. A larger grouped
+  replacement is selected as a whole or not selected, whatever the flag says.
 
 - **Whole-file hunk**: a tracked Hunk with no `@@` text range: a binary change, a mode-only
   (chmod) change, a type change (e.g. file to symlink), or an empty tracked addition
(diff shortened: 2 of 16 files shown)


wkentaro:
> *This was generated by AI — independent final review by a separate Fable agent with no context from the implementation session.*

# Verdict: RECOMMEND-MERGE

## Independent final review

I reviewed this PR against the contract in #225's decided scope and the #232 decision, from the code rather than the PR description, and probed the built CLI live against fresh repositories.

### Contract verification (from source)

- **Guard location and coverage.** The check lives in `_validate_group` in `git_hunk/_lines.py`, exactly where the former one-for-one exemption sat, so `-l` (including the `^` exclusion form), `--include-matching`, and `--exclude-matching` all pass through it, forward and reverse alike. The only production caller of `filter_hunk_lines` is `_apply_line_filter` in `git_hunk/_cli.py`, which all four commands (`stage`, `unstage`, `discard`, `commit`) funnel through.
- **Fail closed, nothing mutated.** The guard raises before any patch is built or applied; `commit` runs the selection filter before staging anything, so a rejection leaves HEAD, index, and worktree untouched. Exit code is 1 (`CliError` without usage), and the message names the pair's body lines, both remedies, and the override flag — matching #232's error-plus-override decision.
- **Override scope.** `--allow-one-sided` exists on all four commands, is a usage error (exit 2) without a selection mechanism, mirroring `--regex`, and is threaded only into the one-for-one branch — the wider grouped-replacement rule stays a hard error with the flag, confirmed by both unit and e2e tests.
- **Docs truthful.** Core skill, README, CONTEXT.md glossary, help text (`--allow-one-sided` row shared by all four commands' help), and the changelog `[Unreleased]` **Breaking:** entry (#251, folded with the #226 documented-then-enforced history) all describe the implemented behavior. No stale "silently commits a half" wording remains outside the changelog history.
- **Eval harness.** The two adversarial solvers that deliberately produce one-sided halves now opt in via the flag, preserving their graded failure mode; I checked every other eval task's selections (`drop_debug_lines`, `separate_formatter_noise` golden, `split_single_hunk` golden) and each selects both sides of any pair it touches, so no eval path breaks.

### Checks and live probes

- Full suite: **801 passed**. `ruff check`, `ruff format --check`, and `ty check` all clean.
- Nine adversarial probes against fresh repositories, all behaving correctly:
  1. `-l ^2` (exclusion spec selecting the addition side) — rejected, exit 1, repo state byte-identical before/after.
  2. `--include-matching 'url\)$' --regex` matching only the deleted side — rejected.
  3. A hunk containing an untouched pair plus a selected pure addition — guard does **not** over-fire; the addition stages without the flag.
  4. `--exclude-matching` excluding both sides of the pair — pair fully unselected, no flag needed.
  5. A pair where both sides lack a trailing newline — rejection names lines 2-3 correctly (the no-newline marker does not shift body numbering) and mutates nothing.
  6. `discard -l 2 --allow-one-sided` on the deletion half — restores the old line beside the new one, as the semantics define.
  7. `commit` rejection — index left empty, HEAD unmoved (guard fires before the internal staging step).
  8. `-l 2,3,6` across two stacked pairs — errors on the second pair (lines 5-6), not the first.
  9. Same selection with `--allow-one-sided` — applies the full first pair plus exactly the named half of the second.

### Non-blocking observations

- The decided scope suggested echoing the existing message template (`cannot partially select lines N-M`); the implementation uses `cannot select one side of lines N-M`. This is a deliberate, clearer variant that still names the pair and the remedies — an improvement, not a deviation worth blocking.
- `test_unstage_allow_one_sided_smoke` and `test_discard_allow_one_sided_smoke` assert only the primary artifact, not full repo state; the stage/commit variants cover the stronger assertions, so coverage is adequate.

The implementation matches the decided scope precisely, fails closed with nothing mutated, and survived every adversarial scenario I constructed beyond the shipped tests.

PR #260, opened 2026-08-09
docs: state the Duplicate Hunk group invariant and re-qualify v0.3.0

### LINKED ISSUE #240: Docs: the core skill never states the Duplicate Hunk group invariant its chaining rule relies on

## Parent

Surfaced by the independent review of #238, which pinned the invariant this text
fails to state.

## What to build

The `core` skill should state the guarantee that its chaining permission depends
on, instead of leaving it to be inferred.

Today the skill says:

> Committing a complete hunk leaves every other non-`conditional` ID valid, so a
> whole plan of complete hunks chains in one call.

That sentence scopes the guarantee to non-`conditional` IDs and says nothing
about what happens to a Duplicate Hunk group. But the skill's worked example
chains a `commit` ahead of further work in a single call, and the relaxation
shipped in #220 permits chaining complete-hunk ID operations *before* an
operation on a Conditional Hunk ID. That permission rests entirely on a stronger
guarantee: committing a complete hunk that is not a member of a Duplicate Hunk
group leaves that group's Conditional Hunk IDs byte-identical.

The invariant is real and is now pinned by a test (#222 / #238). An agent
reading the shipped text cannot learn it. It is implied by omission at the
worked example and stated nowhere.

Note for whoever picks this up: **issue #222's body asserts that the skill
already tells an agent it may chain this way. That assertion was wrong.** It was
written from the intended behaviour rather than from the shipped text. Do not
treat it as evidence.

## Why this is not `ready-for-agent`

The wording is easy; the cost decision is not. Every byte of `core/SKILL.md` is
loaded into an agent's context before its first inspection call and re-read each
turn, and #220 measured and deliberately spent effort shrinking this file. Adding
a sentence is a real per-call token cost, and where it goes matters: #220 found
that moving Hunk ID nuance out of the core loop and into the section where it
applies cut output tokens from 3374 to 2399, because agents stopped reasoning
about a rule most plans never hit.

So this needs a judgement about whether to extend the existing sentence, add one
near the worked example, or leave the guarantee implicit and accept that an agent
must infer it. That is the maintainer's call, and it should be made against a
fresh eval run rather than in the abstract — the #220 baseline is already stale
because #226 modified this file.

## Acceptance criteria

- [ ] The shipped `core` skill states, or deliberately and explicitly declines to
      state, that committing a complete non-member hunk leaves a Duplicate Hunk
      group's Conditional Hunk IDs valid.
- [ ] If the text changes, an eval run confirms the tool-call and token gains
      from #220 survive the addition.
- [ ] The decision and its reasoning are recorded, so the next reader does not
      re-open it from scratch.

## Blocked by

- None — can start immediately, but the eval criterion means it should not be
  measured while #237 is in flight, since that PR changes the eval runner.

wkentaro:
> *This was generated by AI.*

Closes #240.

## Why this is in v0.3.0

#240 was on the v0.3.0 milestone and edits a bundled skill, so it could not
land quietly after the tag: under ADR 0004, any change to either bundled skill
makes the qualifying eval result stale. The maintainer chose to land it and
re-qualify rather than defer it, so this PR carries both the text change and
the new qualifying run.

## The wording judgement

#240's body flagged that the wording is easy but the placement is a cost
decision: every byte of `core/SKILL.md` loads before an agent's first call and
re-reads each turn, and #220 cut output tokens from 3374 to 2399 by moving Hunk
ID nuance out of the core loop into the section where it applies.

The sentence therefore goes in **Splitting and cleanup**, beside the existing
Conditional Hunk ID paragraph, and the plan-and-chain rule in the core loop is
untouched. Cost: +138 bytes, in a section an agent reads only when it meets a
Duplicate Hunk group.

## Acceptance

- [x] The shipped `core` skill states that committing a complete non-member
      hunk leaves a Duplicate Hunk group's Conditional Hunk IDs valid
- [ ] An eval run confirms #220's tool-call and token gains survive the
      addition — pending, compared against the N=3 baseline below
- [x] The decision and its reasoning are recorded (here and in the commit)

## Baseline to beat

The current published N=3 run at `328392e`: git-hunk **8/8 · 25c [25-27] ·
33t [33-35]**. The re-qualification must hold that line and pass every repeat.

### DIFF
--- a/git_hunk/skills/core/SKILL.md
+++ b/git_hunk/skills/core/SKILL.md
@@ -74,6 +74,9 @@ member) renumbers its group. After either, only a bare Repository path may
 follow in that call, as above: `-l` and content matching need a single-hunk
 selection, so they need a fresh `git-hunk list`.
 
+An operation on a complete hunk outside the group leaves those `conditional`
+IDs byte-identical, so it may precede one in the same call.
+
 `--include-matching` and `--exclude-matching` match changed-line content as a
 literal substring; `--regex` switches to a regular expression. Each is
 repeatable and OR'd, but choose only one of `-l`, `--include-matching`, or


wkentaro:
> *This was generated by AI.*

Closing unmerged. The N=3 re-qualification this PR was waiting on did not confirm
#240's acceptance criterion — median tool calls 25 -> 29, turns 33 -> 37 against
the published baseline — so the maintainer chose to leave the guarantee implicit
and keep `core/SKILL.md` unchanged.

The run was not wasted: it is the fresh-eval evidence #240 asked for, and the
decision and full numbers are recorded in
[the resolution comment](https://github.com/wkentaro/git-hunk/issues/240#issuecomment-5232488407).
#240 is closed.

v0.3.0 therefore ships the skill text that the published qualifying run at
`328392e` actually evaluated, and `main` needs no re-qualification.

PR #272, opened 2026-08-17
test: pin that a staged hunk's position counts only unstaged peers

### LINKED ISSUE #246: Test: pin that a staged Hunk's position counts only unstaged peers

## Parent

Surfaced by the independent review of #245, and confirmed there by mutation.

## What to build

A test that fails when the worktree-position calculation for a **staged** Hunk stops
restricting the peers it counts to the **unstaged** ones.

Ordering a Duplicate Hunk group's members adds the net line delta of the peers ahead of a
staged member. Two conditions decide which peers count: the peer must be unstaged, and it
must have the same Repository path. #245 pinned the Repository-path half. The unstaged
half is still uncovered — mutating the guard so that staged peers get counted too leaves
**all 746 tests green**, verified during that review.

The consequence, if it regressed, is the same class of bug #245 closed: a Duplicate Hunk
group's Conditional Hunk IDs shifting because of a change that should not affect them.
Here the culprit would be another *staged* Hunk ahead of the member rather than an
unstaged one under a different path.

Reaching it needs a scenario with at least two staged Hunks in one file — one a group
member, one ahead of it with a net line delta — which `git-hunk stage` can build up
across two calls.

## Acceptance criteria

- [ ] A test builds a Duplicate Hunk group with a staged member, plus a second staged
      Hunk ahead of it in the same Repository path whose net line delta would move the
      member if counted.
- [ ] The test asserts on full Hunk IDs from `--json`, not on human-output prefixes, so a
      prefix-length change cannot mask a renumbering.
- [ ] Dropping the unstaged condition from the peer walk makes the test fail. Verify this
      by making the mutation, observing the failure, and reverting.
- [ ] It lives beside the existing duplicate-identity coverage and follows the suite's
      `_scrubbed_git_env` conventions.

## Blocked by

- None — can start immediately. It does not depend on #245, though the test it adds will
  read naturally beside that one.

wkentaro:
Closes #246.

Ordering a Duplicate Hunk group's members adds the net line delta of the Hunks ahead of a staged member, and two conditions decide which peers contribute: the peer must be unstaged, and it must share the Repository path. #245 pinned the path half. The unstaged half was uncovered — letting staged peers count too left the whole suite green.

## The scenario

A staged peer ahead of the *later* member only pushes it further from its twin, so the peer has to sit ahead of the **earlier** member and be staged along with it. That needs room ahead of the first block, which the shared `duplicate_hunks` fixture does not leave — its file starts at the block, so a top-of-file change would merge into the first member and dissolve the group. Hence the prefixed arrange helper, extracted in the first commit from the sibling test that already built it inline; the new test then reuses that sibling's exact shape: a 100-line insertion, which exceeds the members' separation by enough margin that a double-counted delta actually reorders them.

## Mutation evidence

Each mutation applied to the guard in `_compute_worktree_position`, run against `duplicate_test.py`, and reverted. Each half of the guard now has exactly one killing test, both in that file.

| mutation | result |
| --- | --- |
| `if peer.file != hunk.file:` (unstaged condition dropped) | killed by this PR's test, alone |
| `if peer.status != "unstaged":` (path condition dropped) | killed by #245's test, alone |

The pin is one-directional by design: deleting the peer walk entirely leaves this test green, because it catches over-counting rather than under-counting. `test_duplicate_member_keeps_id_across_different_diff_coordinates` covers the other direction.

No `CHANGELOG.md` entry: `tests/` is repository-only tooling per AGENTS.md.

The issue's acceptance criteria name a `_scrubbed_git_env` fixture; the real one is `_scrubbed_process_env` (`tests/conftest.py`), autouse and inherited via the standard `cli` fixture.

## Test plan

- [x] `make test` — 811 passed
- [x] `make lint` — clean
- [x] Mutation matrix above, with `git_hunk/_hunk.py` verified byte-identical afterwards
- [x] First commit verified alone in a detached worktree

### DIFF
--- a/tests/e2e/hunk_identity/duplicate_test.py
+++ b/tests/e2e/hunk_identity/duplicate_test.py
@@ -7,6 +7,21 @@ from git_hunk._hunk import parse_hunk_range
 from ..conftest import GitHunkCLI
 
 
+def _commit_prefixed_duplicate_blocks(cli: GitHunkCLI) -> list[str]:
+    # Room ahead of the first block, which the shared duplicate-group fixture
+    # does not leave: its file starts at the block, so a change at the top falls
+    # inside the first member's context and merges into it, dissolving the
+    # group. Callers make their own edit on top of the returned lines.
+    prefix = [f"prefix {number}" for number in range(10)]
+    block = ["A", "B", "C", "target", "D", "E", "F"]
+    separator = [f"separator {number}" for number in range(30)]
+    original = prefix + block + separator + block
+    cli.repo.write_file("f.txt", "\n".join(original) + "\n")
+    cli.repo.git("add", "f.txt")
+    cli.repo.git("commit", "-m", "init")
+    return original
+
+
 def test_duplicate_hunks_have_unique_conditional_ids(
     duplicate_hunks: GitHunkCLI,
 ) -> None:
@@ -150,13 +165,7 @@ def test_staged_duplicate_member_ignores_unstaged_hunks_in_other_files(
 def test_duplicate_member_keeps_id_across_different_diff_coordinates(
     cli: GitHunkCLI, ordinal: int
 ) -> None:
-    prefix = [f"prefix {number}" for number in range(10)]
-    block = ["A", "B", "C", "target", "D", "E", "F"]
-    separator = [f"separator {number}" for number in range(30)]
-    original = prefix + block + separator + block
-    cli.repo.write_file("f.txt", "\n".join(original) + "\n")
-    cli.repo.git("add", "f.txt")
-    cli.repo.git("commit", "-m", "init")
+    original = _commit_prefixed_duplicate_blocks(cli)
     insertion = [f"inserted {number}" for number in range(100)]
     changed = insertion + [line.replace("target", "TARGET") for line in original]
     cli.repo.write_file("f.txt", "\n".join(changed) + "\n")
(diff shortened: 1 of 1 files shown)


wkentaro:

Full finalize for this own PR: no rebase needed (`main` hasn't advanced past the merge-base), so the earlier green CI still reflects the merged result.

Ran a report-only two-axis `/code-review` (Standards + Spec) as the safety gate before deciding whether to escalate to `/review-fix`:

- **Standards**: no documented-standard violations. One minor naming nit (`_commit_prefixed_duplicate_blocks` doesn't name its return value in the name) — a judgement call, not a blocker.
- **Spec** (issue #246): all four acceptance criteria verified, including mutation-testing the `_compute_worktree_position` guard directly (dropping the `unstaged` condition does make the new test fail, confirming it isn't vacuous). One scope note: the second commit refactors a pre-existing test to share the new arrange helper, technically beyond what #246 asked for, but it's a same-file, single-purpose dedup with good commit sequencing (refactor lands before the test that uses it), not a blocker.

No meaningful findings surfaced, so `/review-fix` was skipped as unnecessary. `/recommit` found the existing two-commit sequence already optimal (behavior-preserving prep before the test that uses it) — nothing to reshape. `/verify` skipped: test-only change, no runtime behavior touched.

CI is green across all matrix legs (lint, ubuntu/macos/windows × 3.10/3.14), merge state is clean.

**Verdict: recommend-merge.**

PR #273, opened 2026-08-17
test: cover parse_hunk_range directly, including omitted and zero counts

### LINKED ISSUE #247: Test: cover parse_hunk_range directly, including omitted and zero counts

## Parent

Surfaced by the independent review of #245.

## What to build

Direct test coverage for `parse_hunk_range`, the function that turns a bare `@@` header
into its old and new start and count.

It has no dedicated test anywhere in the suite. Every existing check reaches it only
indirectly, through code that happens to call it. #245 also imports it into a test to
decode a Hunk's position, which makes it that test's oracle as well as its subject's
dependency.

That coupling was measured during review and is bounded rather than dangerous: with the
parser deliberately swapping old and new, three sibling tests in
`tests/e2e/hunk_identity/duplicate_test.py` fail loudly, so a parser regression cannot
reach production silently. This ticket is therefore about closing a plain coverage gap,
not about repairing a broken guarantee.

The interesting cases are the ones the header grammar allows but the common path never
exercises: an omitted count (`@@ -1 +1 @@`, which means a count of one), a zero count on
either side, and a header the regex should reject rather than misparse.

## Acceptance criteria

- [ ] A unit test covers `parse_hunk_range` directly, including headers with omitted
      counts on either side and a zero count.
- [ ] A malformed header raises rather than returning a silently wrong range.
- [ ] The test lives beside the other `git_hunk._hunk` unit tests and follows their
      existing style.

## Blocked by

wkentaro:
Closes #247

`parse_hunk_range` had no direct test — every existing check reached it through callers. This pins the header grammar the common path never exercises: an omitted count (meaning 1), a real zero count, the trailing heading kept as suffix, and five malformed headers that must raise.

One non-obvious bit: the zero cases guard the omitted-count default. An omitted count captures `None` and falls back to 1, while `-0,0` captures the truthy string `"0"`, so zero survives; the neighbouring spelling `int(...) or 1` would silently turn a creation's or deletion's `0,0` side into a count of 1.

No changelog entry: repository-only tooling (`tests/`).

## Test plan

- [x] `uv run pytest tests/` — 819 passed

- [x] `make lint` — clean

### DIFF
--- a/tests/unit/_hunk/parse_hunk_range_test.py
+++ b/tests/unit/_hunk/parse_hunk_range_test.py
@@ -1,0 +1,38 @@
+import pytest
+
+from git_hunk._hunk import HunkRange
+from git_hunk._hunk import parse_hunk_range
+
+
+def test_parses_both_counts() -> None:
+    assert parse_hunk_range("@@ -1,3 +4,5 @@") == HunkRange(1, 3, 4, 5)
+
+
+def test_omitted_count_means_one() -> None:
+    assert parse_hunk_range("@@ -1 +4 @@") == HunkRange(1, 1, 4, 1)
+
+
+def test_zero_count_stays_zero() -> None:
+    # git writes a 0,0 side for a created or deleted file; zero is a real
+    # count, not the 1 an omitted count defaults to.
+    assert parse_hunk_range("@@ -0,0 +1,3 @@") == HunkRange(0, 0, 1, 3)
+    assert parse_hunk_range("@@ -1,3 +0,0 @@") == HunkRange(1, 3, 0, 0)
+
+
+def test_keeps_git_heading_as_suffix() -> None:
+    assert parse_hunk_range("@@ -1,3 +1,3 @@ def foo():").suffix == " def foo():"
+
+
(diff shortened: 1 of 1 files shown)


wkentaro:

Full finalize for this own PR: no rebase needed (`main` hasn't advanced past the merge-base), so the existing green CI still reflects the merged result.

Ran the `/code-review` gate (Standards + Spec, parallel sub-agents) as a safety net before deciding whether to escalate to `/review-fix`:

- **Standards**: clean. File naming, `ids=` on all three `parametrize` blocks, and direct-construction style all match sibling tests under `tests/unit/_hunk/`. No baseline-smell flags.
- **Spec** (issue #247): all three acceptance criteria verified against the actual `parse_hunk_range`/`HunkRange` source on `main` — omitted-count, zero-count, and malformed-header cases all check out, no scope creep.

No meaningful findings, so `/review-fix` was skipped as unnecessary. Already a single clean conventional commit, so `/recommit` was a no-op. Test-only change, so `/verify` doesn't apply. CI is green across lint and all OS/Python matrix legs.

Recommending merge.

Issue #249, opened 2026-08-09
Test: pin that a same-start peer counts only when it is a pure insertion


wkentaro:
> *This was generated by AI during review.*

## Parent

Surfaced by the independent review of #245, and confirmed there and again by mutation.

## What to build

A test that fails when the tie-break in the staged-Hunk position walk stops requiring an
exact-start insertion.

Ordering a Duplicate Hunk group's members adds the net line delta of the unstaged peers
ahead of a staged member. "Ahead" is decided by two clauses: the peer starts strictly
before the member, **or** it starts at the same line and is a pure insertion. That second
clause is what lets an insertion landing exactly on the member's start count as preceding
it, while a replacement at the same start does not.

Widening the first clause so it swallows the second — counting any peer whose start is
less than *or equal to* the member's — leaves **all 746 tests green**. Nothing pins the
distinction, so a change that made a same-start replacement count as preceding would ship
silently and shift a staged member's position by that peer's delta.

## Acceptance criteria

- [ ] A test distinguishes a same-start pure insertion, which must count toward a staged
      member's position, from a same-start replacement, which must not.
- [ ] Widening the comparison to `<=` makes the test fail. Verify by making the mutation,
      observing the failure, and reverting.
- [ ] The test asserts on full Hunk IDs from `--json`, or drives the ordering directly at
      the unit layer, rather than on human-output prefixes.
- [ ] It follows the suite's `_scrubbed_git_env` conventions.

## Notes

The cheapest route is likely a direct unit test rather than an end-to-end scenario, since
the behaviour is a pure function of hunk headers and statuses. `tests/unit/_hunk/id_test.py`
already drives this machinery with hand-built Hunks.

This and its sibling issue cover two independent gaps in the same function and would
reasonably be closed together in one change.

## Blocked by

- None — can start immediately.

wkentaro:
> *This was generated by AI during review.*

Sibling gap in the same function: #250. Both are direct unit-test candidates and would reasonably be closed together.

PR #274, opened 2026-08-17
test: pin that a peer's position is read from its pre-image side

### LINKED ISSUE #250: Test: pin that a peer's position is read from its pre-image side

## Parent

Surfaced by the independent review of #245, and confirmed there and again by mutation.

## What to build

A test that fails when a peer's position in the staged-Hunk position walk is read from the
wrong side of its diff.

Ordering a Duplicate Hunk group's members compares each unstaged peer's **pre-image**
start against the staged member's position, because the member's own position is still
expressed in the index's coordinates — the peer's post-image start already includes the
shift the walk is trying to compute.

Reading the peer's start from the post-image side instead leaves **all 746 tests green**.
Nothing pins which side is correct, so the two coincide in every scenario the suite
currently builds.

They only diverge when an earlier unstaged peer has already shifted the coordinates — for
example a peer that inserts lines ahead of the one being compared, so its own pre-image
and post-image starts differ. A test needs at least two unstaged peers ahead of the staged
member, the first with a nonzero net delta.

## Acceptance criteria

- [ ] A test builds a staged Duplicate Hunk group member with at least two unstaged peers
      ahead of it under the same Repository path, the first having a nonzero net line
      delta so the later peer's pre-image and post-image starts differ.
- [ ] Reading the peer's start from the post-image side makes the test fail. Verify by
      making the mutation, observing the failure, and reverting.
- [ ] The test asserts on full Hunk IDs from `--json`, or drives the ordering directly at
      the unit layer, rather than on human-output prefixes.
- [ ] It follows the suite's `_scrubbed_git_env` conventions.

## Notes

The cheapest route is likely a direct unit test rather than an end-to-end scenario, since
the behaviour is a pure function of hunk headers and statuses. `tests/unit/_hunk/id_test.py`
already drives this machinery with hand-built Hunks.

This and its sibling issue cover two independent gaps in the same function and would
reasonably be closed together in one change.

## Blocked by

wkentaro:
Closes #250.

Ordering a Duplicate Hunk group's members compares each unstaged peer's pre-image start against the staged member's position, because that member's own position is still in index coordinates. Reading the post-image start instead left all 811 tests green — nothing pinned which side is correct.

The two sides only diverge once an earlier unstaged peer has already shifted the coordinates, so the fixture puts two unstaged peers ahead of the staged member: the first adds 10 lines, which makes the second straddle the staged member's start (pre-image 20, post-image 30). The staged member lands at worktree line 40, behind its group peer at 37; the post-image reading puts it at 35, ahead.

The expected ordinals are spelled out rather than read back from a second assignment run, because an oracle that shares the sort cancels out the ordering under test — a version that derived them from a second `assign_hunk_ids` call stayed green under both a stubbed-out position walk and a reversed sort.

`CONTEXT.md` gains the matching invariant bullet, alongside the walk's other documented clauses.

## Test plan

- [x] `uv run pytest tests/` — 811 passed

- [x] `make lint` — clean

- [x] Mutation-proven, each mutant run in a throwaway clone: reading the peer's post-image start fails only this test (1 failed, 810 passed); the position walk returning `0` fails it; reversing the group sort fails it.

## Notes for review

Sibling issue #249 covers the function's other gap and is assigned separately, so it is not bundled here. That branch pins its gap with a direct position assertion; this one asserts on Hunk IDs because #250's acceptance criterion asks for the ordering, which is the observable behaviour. Worth reconciling the two styles once both land.

### DIFF
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -138,4 +138,6 @@
   the group are committed.
 - Ordering a Duplicate Hunk group's members counts only the unstaged Hunks with the
   same Repository path, so an unstaged change elsewhere cannot renumber the group.
+- Ordering a Duplicate Hunk group's members places each unstaged Hunk by its pre-image
+  start, because a staged member's own position is still in index coordinates.
 - Each Hunk ID in an inventory addresses exactly one Hunk.
--- a/tests/unit/_hunk/id_test.py
+++ b/tests/unit/_hunk/id_test.py
@@ -3,6 +3,7 @@ import re
 from git_hunk._hunk import Hunk
 from git_hunk._hunk import _compute_text_hunk_id
 from git_hunk._hunk import _compute_whole_file_hunk_id
+from git_hunk._hunk import _hash_id
 from git_hunk._hunk import assign_hunk_ids
 from git_hunk._hunk import count_changes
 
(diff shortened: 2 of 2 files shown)


wkentaro:

Full finalize for this own PR: no rebase needed (merge-base is current `main` tip, 0 commits of divergence). Ran a medium-effort `/code-review` gate: clean, no meaningful findings (one minor wording nit in the new CONTEXT.md bullet, not worth a fix-up — it reads correctly in context of the preceding bullet but could be misread in isolation). Skipped `/review-fix`. Single commit was already clean, so `/recommit` was a no-op. Skipped `/verify` (test/doc-only change, no runtime behavior touched).

Verified the new test is not vacuous: hand-computed the expected worktree positions (37 vs 40) and confirmed via mutation testing that reverting `_compute_worktree_position` to read the post-image start instead of the pre-image start makes this test fail while the rest of the suite stays green.

CI is green (7/7 checks) and untouched this tick, so it is authoritative.

**Verdict: recommend-merge**

PR #275, opened 2026-08-17
fix: read the diff with color disabled

### LINKED ISSUE #271: color.ui=always makes list report no hunks and mutations fail
`git-hunk` parses `git diff` output, but the diff commands in `git_hunk/_git.py` do not pass `--no-color`. When the repository (or the user's global config) sets `color.ui = always` or `color.diff = always`, git emits ANSI escapes even though stdout is a pipe, the parse finds nothing, and `git-hunk` reports an empty inventory on a dirty tree.

The `list` case is the dangerous one: it exits `0` and prints `No hunks.`, so an agent following the core skill concludes there is nothing to commit and stops. That is the same silent-success shape as the subdirectory bug fixed in #194.

## Reproduction

```console
$ mkdir /tmp/colortest && cd /tmp/colortest
$ git init -qb main
$ printf 'a\nb\nc\n' > f.txt && git add f.txt && git commit -qm init
$ printf 'a\nCHANGED\nc\n' > f.txt

$ git-hunk list
unstaged:
f.txt
  88b775a  @@ -1,3 +1,3 @@  +1 -1

$ git config color.ui always
$ git-hunk list
No hunks.                      # exit 0, but the tree is dirty

$ git status --short
 M f.txt
```

`git config color.diff always` alone reproduces it too.

## Affected commands

- `list` — prints `No hunks.` and exits `0`. Silent.
- `stage` — `error: no changed file matches 'f.txt'`, exits `1`.
- `discard` — prints the same `error: no changed file matches 'f.txt'` but **exits `0`**, which looks like a second bug worth checking independently of the color handling.

## Suggested fix

Add `--no-color` to the diff invocations in `git_hunk/_git.py` (`get_diff`, `get_unsupported_changes`), or pass `-c color.ui=false` on every `run_git` call so no future git invocation can inherit a user's color config. This mirrors how `--no-relative` and the explicit `--src-prefix`/`--dst-prefix` already defend the parse against user configuration.

A regression test would set `color.ui=always` in the fixture repository and assert the inventory is unchanged.

## Environment

- git-hunk 0.3.0
- git 2.51.0, macOS

Found while recording a demo where the repository had `color.ui = always` set to get colored `git log -p` output.

wkentaro:
`git-hunk` parses patch-form `git diff` output, but Git emits ANSI escapes even into a pipe when `color.ui` or `color.diff` is set to `always`. Those escapes made `list` silently report `No hunks.` on a dirty tree and left mutation commands unable to resolve Repository paths or Hunk IDs.

`get_diff` now requests `--no-color` at the parse boundary. The separate `--name-status -z` query is unchanged because Git does not color that machine-readable format, even under forced color settings; keeping the override local also avoids exporting a global config override into commit hooks.

Closes #271

## Test plan

- [x] `tests/e2e/color_config_test.py` verifies `list`, `show`, and `list --json` are unchanged under `color.ui=always` and `color.diff=always`
- [x] the `repository_path` matrix runs stage, unstage, discard, and commit under both forced-color settings
- [x] counterfactual check: removing `--no-color` reproduces `No hunks.` and fails the regression coverage
- [x] `uv run pytest -q` — 822 passed
- [x] `make lint`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 
 ## [Unreleased]
 
+### Fixed
+
+- Read the diff with color disabled, so `color.ui` or `color.diff` set to `always` no longer hides every tracked hunk — `list` reported `No hunks.` on a dirty tree, and mutation commands could not resolve their targets (#275).
+
 ## [0.3.0] - 2026-08-10
 
 ### Added
--- a/git_hunk/_git.py
+++ b/git_hunk/_git.py
@@ -44,8 +44,11 @@ def get_diff(*, worktree_root: str, staged: bool) -> str:
     # --src-prefix/--dst-prefix override diff.noprefix and diff.mnemonicPrefix
     # the same way --default-prefix does, but predate it, so the supported git
     # floor stays at --no-relative's 2.28 instead of --default-prefix's 2.41.
+    # --no-color: a color.diff or color.ui set to always emits ANSI escapes
+    # even into a pipe, and the parse would find nothing.
     args = [
         "diff",
+        "--no-color",
         "--no-relative",
         "--src-prefix=a/",
         "--dst-prefix=b/",
--- a/tests/e2e/color_config_test.py
+++ b/tests/e2e/color_config_test.py
@@ -1,0 +1,26 @@
+import pytest
+
+from .conftest import GitHunkCLI
+
+REPORTS = (["list"], ["show"], ["list", "--json"])
+
+
+# color.ui and color.diff reach git's diff colorization through different
+# config lookups, so each key gets its own case. Both keys at once run in the
+# hostile diff config of the repository_path suite, which also covers the
+# mutation commands.
+@pytest.mark.parametrize("key", ["color.ui", "color.diff"])
+def test_forced_color_leaves_the_reports_unchanged(cli: GitHunkCLI, key: str) -> None:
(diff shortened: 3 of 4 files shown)


wkentaro:

Full finalize for this own PR.

**Rebase:** skipped — merge state was already `CLEAN` and zero commits/files diverged from `main` at the merge-base, so the existing green CI still reflects the merged result.

**Review:** ran a cheap `/code-review` (Standards + Spec axes, no `--fix`) as the safety net before deciding on `/review-fix`. Both axes came back clean:
- Standards — no hard violations; comments follow the "comment the why" rule, `CHANGELOG.md` entry matches convention, new test matches the repo's `parametrize`+`ids=` precedent. Only pre-blessed minor smells noted (the inline arrange-block duplication already called out as an accepted tradeoff for surgical PRs).
- Spec — the issue's core ask (`--no-color` on both `diff` invocations in `git_hunk/_git.py`, plus a regression test) is fully implemented. Mutation-tested: removing `--no-color` from `get_diff` reproduces the issue's exact `No hunks.` symptom and fails all three new parametrized cases. Both technical claims in the new code comments (`color.diff` outranking `color.ui`; a global `-c` leaking into commit hooks) were independently verified empirically. No scope creep beyond a natural extension of the existing hostile-config fixture.

No meaningful findings on either axis, so `/review-fix` was skipped as unnecessary. Commit history is already a single clean commit, so `/recommit` had nothing to reshape. CI is green (lint + all 6 test matrix jobs) with no push made this tick.

**Verdict: recommend-merge.**

PR #289, opened 2026-09-04
test: pin same-start insertion ordering

### LINKED ISSUE #249: Test: pin that a same-start peer counts only when it is a pure insertion

wkentaro:
Closes #249.

The focused test is mutation-proven: widening `<` to `<=` makes its replacement assertion fail, while the real implementation passes.


wkentaro (commented):
Verdict: approve. Verified against a5409e0.

- Hand trace matches: the insertion peer at `@@ -20,0 +21,2 @@` moves the staged member's position to 22, past its group peer at 21, giving ordinal 1. The replacement peer at `@@ -20 +20,3 @@` is rejected by the same-start clause, so the position stays 20 and the member gets ordinal 0.

- Mutation claim reproduced: widening the comparison to `<=` fails the replacement assertion. Dropping the second clause entirely fails the insertion assertion. Both mutants are killed by this one test.

- Acceptance criteria of #249 met: unit layer, full hashed IDs, existing hand-built Hunk helper, no Git process involved.

- No changelog fragment needed for a test-only change. CI green on all seven jobs.

Non-blocking: #249 mentions a sibling issue in the same function that could be closed together. Fine as a standalone change.

wkentaro:

Retrospective review of the exact shipped head `a5409e0` (merged as `816ba4f`, second parent `a5409e0`).

The change is 18 added lines in `tests/unit/_hunk/id_test.py`: one unit test driving `assign_hunk_ids` with hand-built Hunks, no Git process. The hand trace matches production code: the insertion peer `@@ -20,0 +21,2 @@` satisfies the same-start pure-insertion clause and moves the staged member behind its group peer, while the replacement peer `@@ -20 +20,3 @@` fails both clauses and leaves it ahead. Both ordinals are asserted as full conditional IDs, so the test isolates exactly the clause #249 names.

Mutation checks were reproduced independently at `a5409e0`: widening `<` to `<=` fails the replacement assertion; dropping the same-start insertion clause fails the insertion assertion; the unmutated code passes. The full local suite passed with 824 tests. The session-autouse scrubbed-process-environment fixture covers this test.

All #249 acceptance criteria are met. No changelog fragment is required for a test-only change. Lint and all six OS/Python test jobs succeeded at the reviewed head.

Limitation: sibling gap #250 was not folded into this PR, but it was already closed separately and was explicitly non-blocking.

PR #290, opened 2026-09-04
feat: suggest Repository path for cwd miss

### LINKED ISSUE #287: Suggest the Repository path when an operand is relative to the invocation directory
> *This was generated by AI during a Claude/Codex log audit.*

## Problem

The root-relative Repository path contract is intentional (#183/#187), but the generic miss diagnostic does not help an agent correct a natural cwd-relative operand.

A verified Claude Code session ran from a repository's `web/` directory. `git-hunk list` reported `web/app/components/_body.py`, but the agent later passed `app/components/_body.py` to `git-hunk commit` and received:

```text
error: no changed file matches 'app/components/_body.py'

```

The agent had already listed the changes. It retried from the worktree root with `web/app/components/_body.py`, and the same commit succeeded. One deterministic suggestion would have removed that failed call and retry.

## What to change

Keep Repository paths root-relative and keep rejecting the cwd-relative spelling. When a missing non-hex operand becomes one eligible changed Repository path after prefixing the invocation directory's root-relative path, replace the generic tip with:

```text
tip: Repository paths are worktree-root-relative; did you mean 'web/app/components/_body.py'?
```

Do not silently accept the alternate meaning; root-relative stability and the collision rule from #183 remain unchanged.

## Acceptance criteria

- [ ] `stage`, `unstage`, `discard`, and `commit` share the suggestion through the common target-resolution path.
- [ ] A subdirectory invocation suggests the one eligible root-relative path for a cwd-relative miss.
- [ ] The command still exits nonzero and changes nothing.
- [ ] If the typed operand already matches an eligible root-relative path, it keeps winning even when a cwd sibling also exists.
- [ ] If no eligible cwd-relative candidate exists, the current generic `git-hunk list` tip remains.
- [ ] The user-facing diagnostic gets a towncrier fragment.

## Related

- #127 reported the original cwd-relative friction.
- #183 deliberately chose one root-relative meaning.
- #187 implemented that contract.

wkentaro:
Closes #287 by making a failed cwd-relative mutation operand point to its eligible worktree-root-relative Repository path.

The alternate spelling is guidance only: exact Repository-path precedence and atomic failure remain unchanged.


wkentaro:

Reviewed head `0dcad175ae637da336946e5e566dee564f8c8dba`. All seven checks at that head pass: lint, and tests on Ubuntu, macOS, and Windows with Python 3.10 and 3.14. The local suite also passed with 828 tests, along with `make lint`.

The change matches #287. The suggestion lives in the shared target-resolution path used by `stage`, `unstage`, `discard`, and `commit`. Exact Repository-path matching still runs first, the suggestion only names an eligible path, generic fallback remains for ineligible candidates, and the error occurs before mutation. Existing collision coverage verifies that a typed root-relative path wins over a cwd sibling. The `changelog.d/290.added.md` fragment is correctly numbered.

The Windows repair is verified. The earlier head derived the prefix lexically from Python's cwd and Git's worktree-root spelling; all four new suggestion cases failed on Windows. This head gets both coordinates from one `git rev-parse --show-toplevel --show-prefix` call. The two Windows jobs are green, and root, nested, normalized, colliding-sibling, and generic-fallback behavior were checked directly.

Limitations: Windows behavior was verified through the two green CI jobs rather than a local Windows run. Root-relative precedence relies on the existing collision test. The capitalized `Repository path` wording matches the issue and repository domain terminology.

PR #291, opened 2026-09-04
feat: explain directory operand rejection

### LINKED ISSUE #288: Explain exact-path semantics when an operand names a directory
> *This was generated by AI during a Claude/Codex log audit.*

## Problem

Directory operands intentionally do not expand under the Repository path contract (#183/#187), but they currently fail as if the directory were simply an unknown changed file:

```text
error: no changed file matches 'labelme/translate'

```

Two independent Claude Code sessions made that exact call after changing many tracked translation files. Each retried with shell-expanded exact paths (`labelme/translate/*.ts labelme/translate/*.qm`) and succeeded, committing 64 and 66 hunks respectively. The repeated recovery shows that the platform already supplies the needed expansion; the missing piece is an error that names the contract and the remedy.

## What to change

When a missing non-hex operand resolves to a directory inside the worktree and eligible changed files exist below it, keep rejecting it but report that directory operands are not expanded and advise passing exact Repository paths (a shell-expanded glob is acceptable when it covers the intended tracked files).

Do not add directory-prefix semantics in this issue; that would reopen the coordinate-system decision and introduce new ambiguity. The smallest useful fix is the diagnostic.

## Acceptance criteria

- [ ] `stage`, `unstage`, `discard`, and `commit` share the directory-specific error through the common target-resolution path.
- [ ] A directory containing eligible changed tracked files gets the tailored error and exact-path guidance.
- [ ] The command exits nonzero and changes nothing.
- [ ] A directory with no eligible changed descendants retains the ordinary no-match diagnostic.
- [ ] The message does not suggest `git add`, which would bypass git-hunk's selection guarantees.
- [ ] The user-facing diagnostic gets a towncrier fragment.

## Related

- #21 added exact file operands.
- #183 deliberately excluded directory, glob, and Git pathspec expansion.
- #187 implemented that contract.

wkentaro:
Closes #288 by turning a changed-directory miss into guidance for git-hunk's exact Repository-path contract.

Directory expansion remains out of scope; the command still fails atomically, including for ineligible descendants and directory symlinks that resolve outside the worktree.


wkentaro:

Reviewed head `983a6f7896667cca224a7871587153beb0bbf95b` against #288. All seven checks at that exact head succeeded; local formatting, lint, and the 96 repository-path tests also passed.

The new branch sits in the only non-hex/no-match target-resolution path shared by `stage`, `unstage`, `discard`, and `commit`. It fires only when the operand is an existing directory that resolves inside the worktree and has a descendant eligible for the current command. Otherwise the ordinary no-match error remains. The message names the exact Repository-path contract, permits shell-expanded glob guidance, and never suggests `git add`.

Direct checks covered `.`, nested directories, normalized operands, subdirectory invocation, per-command eligibility, mixed-operand atomicity, and directory symlinks inside and outside the worktree. The new tests cover all four positive command paths, ineligible descendants, and the external-symlink boundary with unchanged-state assertions. The `changelog.d/291.changed.md` fragment is correctly numbered.

Limitations: all-hex directory names retain the pre-existing hunk-ID interpretation, as #288 explicitly scopes the diagnostic to non-hex operands. Windows and macOS coverage comes from CI; the symlink-specific case is skipped on Windows because Git does not track symlinks there.

PR #299, opened 2026-09-18
fix: ignore configured diff drivers

### LINKED ISSUE #298: Configured external diff and textconv drivers can silently hide tracked changes
## Verified defect

At `5e46e0f5dceb76a377dc06bda1421f7cd74ae688`, `get_diff` invokes porcelain `git diff` without disabling external diff commands or text conversion. A configured driver can replace the source patch with arbitrary display output. Both cases below make a dirty tracked file disappear: `git-hunk list --json` exits 0 with `{"schema_version": 2, "hunks": []}`.

## Reproduce

Run against the current checkout's installed CLI in a temporary repository:

```bash
cd "$(mktemp -d)"
git init -q
git config user.name Audit
git config user.email audit@example.invalid
git config core.autocrlf false
printf 'old\n' > a.txt
git add a.txt
git commit -qm init
printf 'new\n' > a.txt
git config diff.external true
git-hunk list --json
# Incorrect: exit 0, empty hunks.
git config --unset diff.external
printf '*.txt diff=audit\n' > .git/info/attributes
git config diff.audit.textconv true
git-hunk list --json
# Incorrect again: exit 0, empty hunks.
```

`true` is a minimal successful driver that emits no output. Real display drivers can emit transformed text rather than the source bytes that the patch engine requires. The two configurations were reproduced separately on the local checkout; the second is not dependent on the first remaining configured.

## Work

Pin machine-consumed diffs to source patches with `--no-ext-diff` and `--no-textconv` in the shared Git adapter. Inspect its callers so inventory and mutation consume the same canonical patch. Keep the existing color, path-prefix, context, and byte-preservation safeguards.

## Acceptance

- [ ] External diff configuration (`diff.external` and `GIT_EXTERNAL_DIFF`) and a `.gitattributes` textconv driver cannot hide the real tracked changes from list/show, including JSON and staged inventory.
- [ ] Selection and stage/unstage/discard operate on the real source bytes under those configurations; a commit cannot report success based on transformed display content.
- [ ] Add regression coverage beside the existing hostile-diff-configuration tests and a towncrier fix fragment.

This is a new reproducible false-clean case, separate from the fixed ANSI-color case #271. Search of existing open and closed issues found no external-diff/textconv ticket. No data-loss claim or performance estimate is made by this report.

wkentaro:
Prevents configured external diff and textconv drivers from making tracked changes appear clean. Closes #298.

Git's porcelain diff honors display drivers unless explicitly disabled. Inventory and mutation share the same adapter, so pinning both opt-outs there keeps every command on canonical source patches.

Before (`git-hunk list --json | jq '.hunks | length'` with a modified tracked file):

```text
diff.external=true: 0
textconv driver:    0
```

After:

```text
diff.external=true: 1
textconv driver:    1
```

Checks: `just lint`; `just test` (843 passed).


wkentaro:
**Verdict: recommend-merge** (head `e24b8ef`, reviewed against `origin/main` and #298). No defects found.

**Verified**

- `get_diff` in `git_hunk/_git.py` is the only patch producer; every list/show/stage/unstage/discard/commit path routes through it, so `--no-ext-diff` and `--no-textconv` there cover all callers.

- The new `tests/e2e/diff_driver_test.py` catches the bug: against a copy of `origin/main` without the fix, all three cases (`diff.external`, `GIT_EXTERNAL_DIFF`, textconv) fail with "No hunks."; on this branch all three pass. The env case is real because the CLI runs in-process, so the monkeypatched variable reaches git after the session-level `GIT_*` scrub.

- `get_unsupported_changes` (`--name-status`) still reports `R100` under both external-diff configurations, so leaving those flags off that call is fine.

- Adjacent configs also behave on this build: attribute-scoped `diff.<driver>.command`, a `-diff` attribute, and `diff.suppressBlankEmpty` all still list and stage the hunk.

- `just lint` passes; `just test` reports 843 passed. `changelog.d/299.fixed.md` follows the fragment README.

**#298 acceptance**: inventory (list/show, JSON, staged), mutation (stage/unstage/discard/commit on real bytes), regression coverage beside the hostile-config tests, and the fix fragment are all met by the test on this branch.

**Not verified**: the pre-fix commit path was not reproduced on main (the test aborts at the first inventory assertion there), and the exact `jq` command in the PR body was not rerun.

